PR #14, opened 2022-08-09
Require Flowzone tests

ab77:
Change-type: patch


ab77:
@klutchell `Tests / Flowzone / Set branch protection rules` will run on merge, so we can keep branch protection rules updated.

ab77:
@balena-ci I self-certify!

PR #28, opened 2022-08-09
Checkout prior to running actions

ab77:
* remove redundant quotes
* begin comments with lowercase
* correct branding /nomenclature

Change-type: patch


ab77:
@balena-ci I self-certify!

PR #86, opened 2022-08-15
(WIP) Node npm package publishing

ab77:
Change-type: patch


.github/workflows/flowzone.yml
```diff
@@ -490,21 +490,47 @@ jobs:
...
+          echo '//${{ inputs.npmjs_repository }}/:_authToken=${{ secrets.NPM_TOKEN }}' > ~/.npmrc
+
+          npm whoami
+
+          branch="$(echo '${{ github.event.pull_request.head.ref }}' | sed 's/\//-/g')"
+
+          version="$(cat < ../VERSION)"
 
-      # TODO: publish to npmjs
+          # TBC: we could suffix with ${{ github.event.pull_request.head.sha }},
```
ab77, line 520: @klutchell @Page- what do we want our draft NPM packages to be called?

klutchell: Use the git hash like you suggested. The docker flow uses the git hash, the branch name, and the pr number as tags.

.github/workflows/flowzone.yml
```diff
@@ -94,7 +94,7 @@ on:
         description: "NPM repository for NodeJS projects, skipped if empty"
         type: string
         required: false
-        default: ${{ github.repository }}
+        default: registry.npmjs.org
```
klutchell, line 97: We should change this input name to npm_registry

.github/workflows/flowzone.yml
```diff
@@ -490,21 +490,47 @@ jobs:
...
       - name: Publish draft tag
         run: |
-          echo "not yet implemented!"
+          echo '//${{ inputs.npmjs_repository }}/:_authToken=${{ secrets.NPM_TOKEN }}' > ~/.npmrc
+
+          npm whoami
+
+          branch="$(echo '${{ github.event.pull_request.head.ref }}' | sed 's/\//-/g')"
+
+          version="$(cat < ../VERSION)"
```
klutchell, line 518: Seems we should get this from package.json if possible

.github/workflows/flowzone.yml
klutchell: Though if version is always required for npm, then maybe with a version prefix, I think that's how resinCI does it?

.github/workflows/flowzone.yml
```diff
@@ -533,8 +559,6 @@ jobs:
         run: |
           echo "not yet implemented!"
 
-      # TODO: publish to npmjs
```
klutchell, line 536: On merge we still need to re-publish with the final version, right?

.github/workflows/flowzone.yml
```diff
@@ -554,7 +578,6 @@ jobs:
...
```
klutchell, line 557: We don't need to support this flow for now, but it's nice to have for users with a "tag-master" worflow rather than finalizing on merge.

.github/workflows/flowzone.yml
klutchell: er, {version}-{sha}` as you said

.github/workflows/flowzone.yml
klutchell: We also don't know that VERSION will be up one relative dir, definitely use `jq -r '.version' package.json`

.github/workflows/flowzone.yml
```diff
@@ -490,21 +490,47 @@ jobs:
...
+          # TBC: we could suffix with ${{ github.event.pull_request.head.sha }},
+          # however that would create a new version for every PR rebase, which is probably
+          # undesirable, even though that is how resinCI currently seems to do it
+
+          npm version --allow-same-version "${version}-${branch}"
+
+          access=restricted
+          if [[ '${{ steps.npm.outputs.private }}' == '' ]]; then
```
klutchell, line 527: Where is the `npm` step id?

.github/workflows/flowzone.yml
ab77: @klutchell we'll need to do something similar with `npm dist-tags ...` as [resinci-deploy](https://github.com/product-os/resinci-deploy/blob/master/lib/npm.ts)

klutchell:
Versionbot updates the version if they have it installed. Have a look at
the version_draft job and how it checks for the versionbot branch.

.github/workflows/flowzone.yml
ab77: how would we clean up all of the SHAs generated by a PR (incl. rebases, etc.), since `${{ github.event.pull_request.head.sha }}` will always be different on a rebase?

ab77:
Yeah, they can be cleaned by doing a simple `npm dist-tag ls | grep -v ... | xargs npm dist-tag rm`, or something similar.

ab77:
> Versionbot updates the version if they have it installed. Have a look at the version_draft job and how it checks for the versionbot branch.

So I am getting it from the `VERSION` file, which is authoritative (I think), no?

ab77:
This is what the NPM draft job leave the registry as:

```
$ npm dist-tag ls
ab77-operational-819db9ac9c3c554c368fe6cdc692af7de2809e60: 0.7.0-ab77-operational-819db9ac9c3c554c368fe6cdc692af7de2809e60
ab77-operational-d4227baf5830ead5890600298517b5cf721990dd: 0.7.0-ab77-operational-d4227baf5830ead5890600298517b5cf721990dd
ab77-operational: 0.7.0-ab77-operational-819db9ac9c3c554c368fe6cdc692af7de2809e60
latest: 0.7.0-ab77-operational-819db9ac9c3c554c368fe6cdc692af7de2809e60
```

klutchell:
If they aren't using versionbot there won't be a version file. For npm
modules we should only read the version from package.json

klutchell:
Note that the package.json does not get bumped in Flowzone cause it's in
the tests subdirectory and versionbot doesn't find it. We could abandon the
option to specify a working-directory, or in this case we could add a
package.json symlink in the root so it gets bumped like a normal project.

ab77:
> add a package.json symlink in the root so it gets bumped like a normal project.

If VB only looks in the root, then this is the best option (I think).

ab77:
> If they aren't using versionbot there won't be a version file. For npm modules we should only read the version from package.json

I am probably missing something obvious, but it this case VB doesn't update package.json at all:
https://github.com/product-os/flowzone/runs/7901933487?check_suite_focus=true#step:4:47

klutchell:
> > If they aren't using versionbot there won't be a version file. For npm modules we should only read the version from package.json
> 
> I am probably missing something obvious, but it this case VB doesn't update package.json at all: https://github.com/product-os/flowzone/runs/7901933487?check_suite_focus=true#step:4:47

@ab77 this is for a couple of reasons:

1. package.json is in a subdirectory `tests` which should be resolved by the link you created, but maybe not
2. [the flowzone repo type](https://github.com/product-os/flowzone/blob/master/repo.yml) is `generic` and vb uses this type to determine which files to find and version

So the bigger issue here is we are testing this flow in a subdirectory and VB only looks at the top level of projects so we aren't really testing correctly.

We could move the test files to the root, or move testing to an external repo, or add unit testing of some kind with fake github event inputs.

klutchell:
@ab77 your last commit reverted all your flow changes :(

.github/workflows/flowzone.yml
```diff
@@ -497,59 +500,139 @@ jobs:
...
+              access=restricted
+          else
+              access=public
+          fi
+
+          package="$(jq -r '.name' package.json)"
+
+          branch="$(echo '${{ github.event.pull_request.head.ref }}' | sed 's/\//-/g')"
+
+          version="$(cat < "${GITHUB_WORKSPACE}/VERSION")"
```
klutchell, line 592: Use package.json here like we do on line 538

.github/workflows/flowzone.yml
```diff
@@ -497,59 +500,139 @@ jobs:
...
```
klutchell, line 618: Use package.json here like we do on line 538

ab77:
> [the flowzone repo type](https://github.com/product-os/flowzone/blob/master/repo.yml) is `generic` and vb uses this type to determine which files to find and version

The more I learn about VersionBot/Versionist, the more I realise these tools are not the best anymore. the fact that we need to specify an additional piece of metadata just to tell the software what to do is worrying, also https://www.conventionalcommits.org ...

.github/workflows/flowzone.yml
```diff
@@ -497,59 +500,136 @@ jobs:
...
+
+          if [[ "$(jq -r '.private' package.json)" =~ true ]]; then
+              access=restricted
+          else
+              access=public
+          fi
+
+          package="$(jq -r '.name' package.json)"
+
+          branch="$(echo '${{ github.event.pull_request.head.ref }}' | sed 's/\//-/g')"
```
klutchell, line 587: A better regex would be `'s/[^[:alnum:]]/-/g'`

.github/workflows/flowzone.yml
```diff
@@ -497,59 +500,136 @@ jobs:
...
```
klutchell, line 613: A better regex would be `'s/[^[:alnum:]]/-/g'`

klutchell:
> > [the flowzone repo type](https://github.com/product-os/flowzone/blob/master/repo.yml) is `generic` and vb uses this type to determine which files to find and version
> 
> The more I learn about VersionBot/Versionist, the more I realise these tools are not the best anymore. the fact that we need to specify an additional piece of metadata just to tell the software what to do is worrying, also https://www.conventionalcommits.org ...

I like [this PR](https://github.com/product-os/flowzone/pull/88) that calls versionist directly, but we still have the same problems. Fortunately we own it and can change it to behave differently, we just risk breaking existing workflows and automations.

ab77:
@klutchell so the draft part worked, but merged failed because versioned source [wasn't found](https://github.com/product-os/flowzone/runs/7958398102?check_suite_focus=true#step:2:8). What should we do here?

klutchell:
I'll have to go and look at the other versioning PR, cause something was merged last week that might have broken things.

I don't think it's related to your node changes but I'll confirm

klutchell:
The issue is that on merge events we need to download the versioned source artifact from the previous workflow run, because the PR branch has already been deleted. For any of the `_merge` jobs we need to use a different artifact download action that searches for artifacts matching that commit on previous runs of the workflow.

https://github.com/product-os/flowzone/blob/9a52689db73af7bfbc9b7e1f46156272e9f49fe8/.github/workflows/flowzone.yml#L365-L378

This is part of the reason I wanted to replace the artifact storage with something else this week. I can patch the `npm_merge` job and start looking at alternatives for persistent artifacts.

klutchell:
I didn't expect to run into the above issue because I thought the merge jobs would not require any source code, since they should just be finalizing existing artifacts. I guess that turned out to be difficult to do with npm and no source?

klutchell:
Resolved by https://github.com/product-os/flowzone/pull/90

PR #125, opened 2022-08-25
Fix input parsing when multiple docker images are provided

klutchell:


klutchell:
@balena-ci I self-certify!

PR #133, opened 2022-08-26
Fix tagging of draft docker images

klutchell:
Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #137, opened 2022-08-26
Add cool logo

phil-d-wilson:
Change-type: patch
Signed-off-by: Phil Wilson <phil@balena.io>


phil-d-wilson:
Not sure if I can self-certify with this flowzone nonsense

klutchell:
> Not sure if I can self-certify with this flowzone nonsense

Self-certify still works as long as resinCI is around and monitoring the repo. It seems to only be broken on the handbook right now.

klutchell (approved):

PR #139, opened 2022-08-27
Remove required status checks input

klutchell:
This input was very easy to break with formatting issues
and did not take into account the fact that Flowzone job
names may change.

A suitable replacement should be considered to adjust the existing
branch rules rather than apply all new ones.

Change-type: minor


klutchell:
@balena-ci I self-certify!

PR #142, opened 2022-08-27
Restore missing balena finalize job

klutchell:
This is nearly identical to the balena publish
job but it makes it clear from the job flow what
is being done.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #144, opened 2022-08-27
Only run protect branch on unmerged pull requests

klutchell:


klutchell:
@balena-ci I self-certify!

PR #150, opened 2022-08-28
Remove node versions input and limit to 16.x for now

klutchell:
This will be replaced by testing a range of LTS releases
defined within the engines.node section of the package.json.

Change-type: minor


klutchell:
@balena-ci I self-certify!

Issue #149, opened 2022-08-28
Use engines field of package.json to limit LTS node versions for testing


klutchell:
We can default to on or more LTS node versions and allow this to be adjusted with the `engines.node` range in `package.json`

PR #151, opened 2022-08-29
Find the intersection of the node engines and recent LTS releases

klutchell:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/149


klutchell:
@balena-ci I self-certify!

PR #152, opened 2022-08-29
Set artifact retention days to 1

klutchell:
We only need artifacts for the lifetime of this workflow
so there is no need for them to live longer than a few minutes.

Change-type: patch


klutchell:
@balena-ci I self-certify!

Issue #146, opened 2022-08-28
Support advanced docker build configuration via docker-compose or bake files


klutchell:
The docker build job is not currently using all of the [available options](https://github.com/marketplace/actions/build-and-push-docker-images#inputs) in the official GitHub Action.

https://github.com/product-os/flowzone/blob/4ed9faf892413f51c9e13b83f6d3006818c8ab6d/.github/workflows/flowzone.yml#L676-L690

Some things we omitted but should support in Flowzone:
- `build-args`
- `file`
- `secrets`
- `target`
- `platforms`
- ...

Originally some of these were exposed as inputs on the workflow, but inputs are often the reason of breaking compatibility so we should use inputs as little as possible and instead infer these settings from the project.

The docker-compose file would have some of these settings if we had a way to know which service name to parse.

Another option would be to support discovery of [a bake file](https://docs.docker.com/build/bake/file-definition/) that would have all the fields we need for advanced Docker build options and would only be required if not using the Flowzone defaults.

PR #155, opened 2022-08-29
Use docker-bake action if a docker-bake.hcl file is found

klutchell:
This allows the use of advanced build options like context, dockerfile,
build args, etc.

Change-type: minor
Resolves: https://github.com/product-os/flowzone/issues/146


klutchell:
@balena-ci I self-certify!

PR #156, opened 2022-08-29
Replace context check with event types parser

klutchell:
This moves the custom project logic into Project Types and
allows a single job to determine the types of events to support.

Change-type: minor


klutchell:
@balena-ci I self-certify!

PR #157, opened 2022-08-29
Fix cases where node engine is not specified

klutchell:
Change-type: patch
See: https://github.com/product-os/flowzone/issues/149
Replaces: https://github.com/product-os/flowzone/pull/151


klutchell:
@balena-ci I self-certify!

PR #158, opened 2022-08-30
Only check node engine versions if npm is enabled

klutchell:


klutchell:
@balena-ci I self-certify!

PR #161, opened 2022-08-30
Fix job conditions when ancestors of all_tests are skipped

klutchell:


klutchell:
@balena-ci I self-certify!

PR #162, opened 2022-08-30
Avoid failure if event object cannot be parsed as JSON

klutchell:
Special characters, especially those used by Renovate, will
fail the toJSON command.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #163, opened 2022-08-30
Remove All Tests and All Published jobs

klutchell:
Instead make sure that Protect Branch always runs so branch
protection rules can depend on it being pass/fail and not skipped.

Change-type: minor


klutchell:
@balena-ci I self-certify!

PR #167, opened 2022-08-31
Provide all secrets and inputs to the composite action via json

klutchell:


klutchell:
@balena-ci I self-certify!

Issue #154, opened 2022-08-29
Document how to skip CI on the README


klutchell:
https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

> Workflows that would otherwise be triggered using `on: push` or `on: pull_request` won't be triggered if you add any of the following strings to the commit message in a push, or the HEAD commit of a pull request:
    -       `[skip ci]` 
    -       `[ci skip]` 
    -       `[no ci]` 
    -       `[skip actions]` 
    -       `[actions skip]`

> Alternatively, you can end the commit message with two empty lines followed by either:
    -       `skip-checks:true` 
    -       `skip-checks: true` 

> You won't be able to merge the pull request if your repository is configured to require specific checks to pass first. To allow the pull request to be merged you can push a new commit to the pull request without the skip instruction in the commit message.

PR #170, opened 2022-09-01
Document how to skip CI on the README #154

fisehara:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/154


klutchell (commented):
Looks good @fisehara, thanks for the contribution!

What made you use `minor` for change type? To me something like documentation could just be a `patch`.

fisehara:
@klutchell I'll change it to patch. Sometimes I feel like adding new stuff, even for doc is not a patch. Anyway, patch is fine 👍

klutchell (approved):

Issue #111, opened 2022-08-24
Branch protection rules are not applied if there are no existing rules


klutchell:
```
Run url='https://api.github.com/repos/klutchell/unbound-docker/branches/main/protection'
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#update-branch-protection"
}
```
https://github.com/product-os/flowzone/blob/e0f2ea727e8d6892e28b9b3ea7b2d38ba8565d2a/.github/workflows/flowzone.yml#L262

Issue #141, opened 2022-08-27
Extend branch protection rules to amend vs replace


klutchell:
The current branch protection rules job will apply some hardcoded values for branch protection.

https://github.com/product-os/flowzone/blob/a4601e64e47fe1684828cb1848cbf7f36c3744a9/.github/workflows/flowzone.yml#L354-L390

However, we should allow custom required checks, reviewers, etc. This could best be solved by getting the current rules from the GitHub API, amending them, and putting them back to the GH endpoint.

1. GET existing protections from GitHub API
2. Convert to [JSON object](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson)
3. Remove ResinCI checks
4. Add Flowzone checks
5. Leave other checks alone
6. Update required reviewers based on input
7. Convert back to [JSON string](https://docs.github.com/en/actions/learn-github-actions/expressions#tojson)
7. PUT back to GitHub API

fisehara:
@klutchell can you please add some details about what:
1. Remove ResinCI checks are

klutchell:
@fisehara 

1. The current resinCI checks can be found on any balena repo currently implementing resinCI. You can test this by removing `.resinci.yml` in this PR which will cause resinCI to start running on this PR again. Then you can see what tests are run, and these are the tests we would want to REMOVE from branch protection rules for repos adopting Flowzone.

2. The Flowzone checks we want to add are currently hardcoded. Ideally we could start producing these dynamically, but for now just the ability to remove any `resinCI / * ` and `Flowzone / *` checks and replace them with the current Flowzone checks would be enough to ensure that other unrelated branch rules are not lost. For example, if a branch rule depends on a Jenkins job we don't want to remove it. Hence the request to PATCH the branch rules vs POST.

We can get a meet today if that would help! LMK!

Issue #140, opened 2022-08-27
Applying branch protection does not catch failures


klutchell:
See https://github.com/product-os/flowzone/issues/111

We need to evaluate the response from the GitHub endpoint and confirm the settings were applied, or fail the step.

PR #171, opened 2022-09-01
Add check of branch protection API call response

fisehara:
Separate branch protection job into steps
1. fetch existing branch protection object
2. parse and remove resinCI and existing flowzone checks
3. prepare new branch protection update object.
4. check result containing `required_approving_review_count`

resolves: https://github.com/product-os/flowzone/issues/111 https://github.com/product-os/flowzone/issues/141 https://github.com/product-os/flowzone/issues/140

Change-type: minor

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -1152,9 +1152,10 @@ jobs:
       run:
         working-directory: .
         shell: bash --noprofile --norc -eo pipefail -x {0}
-
+    env:
+      BRANCH_PROTECTION_URL: ${{ github.api_url }}/repos/${{ github.repository }}/branches/${{ github.event.repository.default_branch }}/protection
     steps:
-      # loop over all needed jobs and fail this job if any of them failed
+      #loop over all needed jobs and fail this job if any of them failed
       - name: Check needed jobs
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/flowzone.yml
```diff
@@ -1120,3 +1120,8 @@ jobs:
           }')"
 
           echo ::set-output name=result::"${result}"
+
+          if [[ `$(jq '.required_pull_request_reviews.required_approving_review_count' <(echo "$result"))` -eq ${ inputs.required_approving_review_count } ]]; then
```
klutchell, line 1124: I don't think you want the backticks here since the $() already executes.

klutchell (commented):
Is there a way we can check for any failure response and solve https://github.com/product-os/flowzone/issues/140 at the same time?

With this change if the approving reviewers count was already up-to-date, but the required checks were not applied we would miss it.

klutchell:
Might also need to consider how this would impact https://github.com/product-os/flowzone/issues/141

fisehara:
@klutchell thanks for the input. I'll work on a step based execution of the branch_protection job.

fisehara:
@klutchell LMK if the amount of checks is sufficient for now.

klutchell (approved):
Looks great, thanks @fisehara !

Issue #172, opened 2022-09-01
Generating CHANGELOG.yml fails when no versioned PRs have merged


mehalter:
I am setting up a new repository (very similar to https://github.com/balenaltd/handbook) It has never been versioned before and this is the first PR which we are hoping jump starts the versioning. It looks like flowzone is failing to generate the changelog probably due to the fact that it has never been "released" or versioned prior to the pull request

https://github.com/balenaltd/security/pull/1

The relevant logs of the error can be found here: https://github.com/balenaltd/security/runs/8142755809?check_suite_focus=true

klutchell:
```
+ ./generate-changelog.sh .
Generating CHANGELOG.yml for `balenaltd/security`
Generating CHANGELOG.yml...
mv: cannot stat '/tmp/tmp.vD3pNx9o6K/v*.yml': No such file or directory
```
This script doesn't handle situations where the repo has never had a versioned PR merged.
https://github.com/product-os/versionist/blob/master/scripts/generate-changelog.sh

Versionist needs a lot of work, including fixing this script, but for now we can have Flowzone ignore this failure and get the new version via some fallback method.

PR #173, opened 2022-09-01
Continue when the changelog cannot be generated

klutchell:
This can happen if the repo has never have a versioned PR merged.

Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/172

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -343,8 +343,8 @@ jobs:
         shell: bash --noprofile --norc -eo pipefail -x {0}
 
     outputs:
-      new_tag: ${{ steps.versionist.outputs.new_tag }}
-      version: ${{ steps.versionist.outputs.version }}
+      new_tag: ${{ steps.new_version.outputs.tag }}
+      version: ${{ steps.new_version.outputs.semver }}
 
     steps:
       - name: Checkout source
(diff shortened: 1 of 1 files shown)


mehalter:
@klutchell this looks great! I went through the code and it looks reasonable (given my unfamiliarity with the full code base). Is there a way for me to tell my PR to use this branch to test it before I put any sort of "approval" review on it?

klutchell:
Don't worry about the approval yet, I'm going to re-review with fresh eyes
today. I also have some questions about your PR.

To test this PR on your PR, you can change ***@***.*** to a
ref from my PR, like the branch or commit.

mehalter:
Ah thanks for the reply @klutchell I noticed that and have been giving it a try :) I think it's running, I ran into a new issue but it is definitely on my end haha I set the version in the `package.json` to the initial version I wanted `0.1.0` and it was erroring because the "current version" is the same as the "updated version" hehe

mehalter:
https://github.com/balenaltd/security/pull/1

Looks like i got passed that hurdle though! Also could you disable versionbot on the balenaltd/security repo for me? :)

mehalter:
Oh @klutchell one thing i did come across, in the workflow file for Flowzone the steps for parsing the old version and parsing the new version are both named `Parse version`. It might be helpful/more clear to specify that one is parsing the old version and one is parsing the new :)

klutchell:
> It might be helpful/more clear to specify that one is parsing the old version and one is parsing the new :)

Good call @mehalter , in fact I'm going to change the first version check to look for existing tags (since that's what versionist does) and we can avoid cases where the initial version was set in package.json but there is no tag.

klutchell:
@mehalter can you retry this branch with your PR again? I changed the logic to use tags for the old version, and since your repo doesn't have any versioned tags yet it shouldn't matter what is in package.json (it will be overwritten by versionist anyway).

mehalter:
@klutchell looks like it worked perfectly!

klutchell:
@balena-ci I self-certify!

Issue #180, opened 2022-09-03
Support custom closed/cleanup actions for closed PRs.


klutchell:
There are some project types that require a cleanup when PRs are closed (with or without merging).

For example, when [deploying drafts to GH pages](https://github.com/marketplace/actions/deploy-pr-preview) we want to delete the extra branches/directories when a PR is closed.

PR #181, opened 2022-09-05
Add support for custom clean actions

klutchell:
These composite actions will run when a PR is closed without merging.

Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/180


klutchell:
@balena-ci I self-certify!

PR #183, opened 2022-09-05
Allow clean jobs to run when a PR is merged

klutchell:
Composite jobs can check additional conditions like whether
or not the PR was merged.

Change-type: patch


klutchell:
@balena-ci I self-certify!

Issue #179, opened 2022-09-02
Consider adding concurrency groups


klutchell:
https://docs.github.com/en/actions/using-jobs/using-concurrency

PR #184, opened 2022-09-05
Cancel in progress jobs of the same ref

klutchell:
Cancel parallel flowzone runs from the same ref that may
cause conflicts.

Change-type: minor
Resolves: https://github.com/product-os/flowzone/issues/179


klutchell:
@balena-ci I self-certify!

PR #187, opened 2022-09-08
Install dependencies before npm publish if there are relevant scripts

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #191, opened 2022-09-08
Use npm pack to artifact the tested npm project files

klutchell:
This will prevent the publish and finalize steps from installing packages with secrets access, and ensure that projects without a package.json won't end up with different dependency versions when publishing.

Change-type: patch


.github/workflows/flowzone.yml
```diff
@@ -603,38 +607,29 @@ jobs:
...
-              npm ci
-            else
-              npm i
-            fi
-          fi
-
       - name: Publish final release
         run: |
-          npm --loglevel=verbose --logs-max=0 publish --tag "latest"
+          pack="$(ls /tmp/npm-*/*.tgz | tail -n1)"
```
Page-, line 631: Is this guaranteed to find the latest? I don't know if it'll hit issues with using lexical rather than numerical sorting?

klutchell: We can add a `| sort -V` to to try and sort more reliably? I didn't want to spend too much time on this part.

Page- (approved):

klutchell:
I just realized that I'm setting the full draft version in package.json during build/test and I'm not updating it to the final version during finalize. I want to avoid unpacking the tarball though...

Issue #188, opened 2022-09-08
Default to not publish a docker container when docker_images are not set explictly


fisehara:
Repositories may define a `docker-compose.yml` for sophisticated testing
The resulting docker images are not meant to be published.

https://github.com/product-os/flowzone/blob/8624dde2c0e893aaf8e16d92ec89494a49d277d1/.github/workflows/flowzone.yml#L755-L762

Page-:
I think this is already supported by setting `docker_images` to empty? (and the default looks to be empty?)

fisehara:
@klutchell @Page- FYI setting a job strategy matrix with an empty array will not avoid running the job.
The matrix.<expandedVariable> will be null and the job fails. 
Thus I'm adding a 
``` yaml
    if: |
      !failure() && !cancelled() &&
      needs.docker_test.result == 'success' &&
      needs.event_types.outputs.do_draft == 'true' &&
      needs.project_types.outputs.docker_images != []
```
Conditional execution to the docker_publish job.

PR #192, opened 2022-09-08
Default docker_images inputs to ''

fisehara:
Skip docker publish by default, when not explicitly set. 

resolves: #188

Change-type: patch


fisehara:
@balena-ci I self-certify!

PR #194, opened 2022-09-08
Fixing docker_images string and add empty check

fisehara:
Fixing docker_images string and add empty check

Change-type: patch


fisehara:
@balena-ci I self-certify!

PR #197, opened 2022-09-12
Test only from docker-compose.test.yml

fisehara:
Run single service sut so that exit code is derived from this service

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -283,7 +283,7 @@ jobs:
       - name: Check for docker-compose.test.yml
         id: docker_compose
         run: |
-          if [ -f docker-compose.test.yml ] && [ -f docker-compose.yml ]; then
+          if [ -f docker-compose.test.yml ]; then
             echo "found docker-compose.test.yml"
             echo ::set-output name=enabled::"true"
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/flowzone.yml
```diff
@@ -742,7 +742,12 @@ jobs:
           if [[ ! -z "${COMPOSE_VARS}" ]]; then
             echo ${COMPOSE_VARS} | base64 --decode > .env
           fi
-          docker compose -f docker-compose.yml -f docker-compose.test.yml up --exit-code-from sut
```
ab77, line 745: @fisehara I think `--exit-code-from sut` is important...

ab77 (changes requested):

.github/workflows/flowzone.yml
fisehara: @ab77 I thought it would be fine to just run the sut service but I just add the `--exit-code-from` to the command but up only the sut service. 

This requires the docker-compose.test.yml to specify a test service with all dependencies defined with `depends_on`

ab77 (approved):

.github/workflows/flowzone.yml
klutchell: @fisehara This change (`up sut` vs `up --exit-code-from sut`) results in the logs from the dependent services being suppressed, and we only see logs from the sut service.

Was there an issue with starting all services and only using the exit code from sut?

It also goes against what was working in resinCI (not that it should matter).
https://github.com/product-os/scripts/blob/652fdd41851cda614a6a4c29cf7a14ce17bc9b6b/docker/build.sh#L238-L242

Issue #198, opened 2022-09-12
feature: allow specifying the environment for `balena push` 


cywang117:
I didn't see anything in the docs about this so I'm assuming it's not yet a feature, but it would be nice if there was at least the option of specifying staging or production with the balena workflow. This would allow balena-os/balena-supervisor repo for instance to fully switch over to Flowzone.

PR #199, opened 2022-09-12
Specify balenaCloud environment

ab77:
Resolves: https://github.com/product-os/flowzone/issues/198


ab77:
@balena-ci I self-certify!

PR #202, opened 2022-09-12
Add Dockerfile parameter

ab77:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -45,6 +45,11 @@ on:
         type: string
         required: false
         default: "."
+      dockerfile:
+        description: "Default Dockerfile to look for."
+        type: string
+        required: false
+        default: "Dockerfile"
       docker_images:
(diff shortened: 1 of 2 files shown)


ab77:
@balena-ci I self-certify!

klutchell:
@ab77 I was trying to avoid extra inputs like this as inputs are an arbitrary interface that is easy to break.

In a previous [commit ](https://github.com/product-os/flowzone/pull/145/commits/5af10a8a4a07017d739eb1fc33545ce9a349e5c1)I removed this input and [replaced it with support for docker bake files](https://github.com/product-os/flowzone/pull/155).

We should remove this input, or note on the README that it is ignored if using docker bake files.

ab77:
OK I see, I'll revert this then cc @pipex @cywang117 @Page-

ab77:
reverted in https://github.com/product-os/flowzone/pull/205

PR #203, opened 2022-09-12
Install system dependencies

ab77:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -543,6 +543,12 @@ jobs:
           echo "::set-output name=sha_tag::${sha_tag}"
           echo "::set-output name=version_tag::${version_tag}"
 
+      - name: Install system dependencies
+        run: |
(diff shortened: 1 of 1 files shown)


ab77:
Can you please try this with balena-supervisor (dbus):

```
jobs:
  flowzone:
    name: Flowzone
    uses: product-os/flowzone/.github/workflows/flowzone.yml@ab77/operational
    ...
```

You'll need to have `package.txt` in the root, which should be piped to `apt ...`

@pipex @cywang117

pipex:
@ab77 is there an advantage to this vs using `preinstall` in `package.json` as Christina did below? If there isn't i'd be inclined to go with the `preinstall` option, as it removes the maintenance burden for this repo

https://github.com/balena-os/balena-supervisor/pull/2010/commits/d0b53558875c3a211262b6a3470a58491ed4c4e6#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R28

PR #205, opened 2022-09-13
Revert Dockerfile parameter

ab77:
* use advanced Docker build flow (.hcl) instead

Change-type: patch


ab77:
@balena-ci I self-certify!

PR #208, opened 2022-09-14
Support specifying custom workflow names for branch protections

Page-:


.github/workflows/flowzone.yml
```diff
@@ -1226,18 +1231,23 @@ jobs:
 
           required_status_checks__strict=$(echo $jsondata | jq ".required_status_checks.strict")
 
+          flowzone_prefix='Flowzone'
+          if [[ '${{ inputs.job_name }}' != 'Flowzone' ]] ; then
+            flowzone_prefix='Flowzone / ${{ inputs.job_name }}'
```
pipex, line 1236: This might be a possible alternative if I'm reading https://docs.github.com/en/actions/learn-github-actions/contexts#job-context correctly

```suggestion
            flowzone_prefix='Flowzone / ${{ jobs[github.job].name }}'
```

Page-: I did consider/try it a bit @pipex but didn't have much luck, however I did give up pretty quick as it's fairly late and was hoping to just get something that works for now and can be improved later. If you want to dive into it more then that'd be great and I'm 100% for it

pipex: Yeah, i figured you were testing an initial approach first for later improvement. I can give it a shot to see if I find the right context to use

Page-:
@balena-ci I self-certify!

PR #209, opened 2022-09-14
Support multiple balena environments to push to

Page-:
Change-type: minor

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -61,7 +61,7 @@ on:
         required: false
         default: ""
       balena_environment:
-        description: "balenaCloud environment"
+        description: "Comma-delimited string of balenaCloud environments"
         type: string
         required: false
         default: balena-cloud.com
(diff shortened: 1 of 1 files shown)


.github/workflows/flowzone.yml
```diff
@@ -1004,10 +1006,11 @@ jobs:
         working-directory: .
         run: tar -xvf source.tar
 
-      - uses: balena-io/deploy-to-balena-action@v0.13.0
+      - name: Deploy to ${{ matrix.balena_env.name }} 
+        uses: balena-io/deploy-to-balena-action@v0.13.0
         with:
-          balena_token: ${{ secrets.BALENA_API_KEY || secrets.BALENA_API_KEY_PUSH }}
-          environment: ${{ inputs.balena_environment }}
+          balena_token: ${{ secrets[matrix.balena_env.secret] }}
```
pipex, line 1012: I haven't been able to figure out why this doesn't work

pipex: This works actually, is just that the flowzone repo uses `BALENA_API_KEY_PUSH` instead of `BALENA_API_KEY`

pipex:
Closing this as using multiple environments starts to get out of scope for flowzone and it requires trading off in UX.

PR #212, opened 2022-09-15
Revert "Support specifying custom workflow names for branch protections"

pipex:
This reverts commit edad4416fdc7f6f6bf9e59ee0a5bcf51f9780d9f.

Although this works, it can potentially cause stale branch protections to become unremovable if the workflow name is changed

Change-type: patch


pipex:
There is some debate on the usefulness of the feature, so closing this for now.

PR #213, opened 2022-09-16
Add npm_test_command as an option for changing the npm test command

thgreasi:
Change-type: minor
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -85,6 +85,11 @@ on:
         type: string
         required: false
         default: "Flowzone"
+      npm_test_command:
+        description: "The command that should be used to start the npm tests"
+        type: string
+        required: false
+        default: "npm test"
 
 # https://docs.github.com/en/actions/using-jobs/using-concurrency
 concurrency:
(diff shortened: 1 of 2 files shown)


klutchell:
IMO we should avoid adding new inputs for every edge case. As soon as a new input is used we can never change or remove it without breaking repos.

What is the use case here? Why can't the behaviour of npm test be changed? Or why not use a custom test stage as documented on the readme?

thgreasi:
This will help migrating the SDK tests sooner than later b/c atm I would need a way to provide a set of different user credentials for each node version, and atm I can't do such a thing.
The short term solution I came up with was adding this to disable the default node tests and add a custom test workflow in the SDK that reads the repo's secrets and starts the tests with those injected.

I heard that the current workaround w/o this is to set `"test": "echo 'Skipping'"` in the package.json, which I didn't really love.

klutchell:
I think adding a fake test to the sdk to avoid failing jobs is preferable
to adding a feature to Flowzone that shouldn't really be required and will
have to be maintained indefinitely.

thgreasi:
> I think adding a fake test to the sdk to avoid failing jobs

In that case it might make sense to wait until flowzone adds support for injecting repo secrets to the test runners, so that `npm test` still does the expected thing.

Page-:
I completely agree with @klutchell here @thgreasi, this command is only desired as a way to have `npm test` but not have flowzone use it which is very specialized to the current SDK needs

PR #214, opened 2022-09-16
Enable publishing/finalizing docker images without docker-compose tests

Page-:
Change-type: minor

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -181,6 +181,7 @@ jobs:
       npm: ${{ steps.npm.outputs.enabled }}
       npm_private: ${{ steps.npm.outputs.private }} # can be null or unset
       docker_compose: ${{ steps.docker_compose.outputs.enabled }}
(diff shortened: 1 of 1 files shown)


klutchell:
Why would we allow publishing without tests? Even just a dummy test would
be worth adding to any repo that's missing one.

Page-:
@klutchell part of the problem is I don't know how I'd even add a dummy test that doesn't actually do anything. That said the primary reason is to allow parity with balenaCI for repos where they had balenaCI publishing but still used eg circleCI testing so it's not that they don't have tests but rather that those tests are not expected to be run via flowzone (yet)

Page-:
Also the other issue is that the repos were already converted to flowzone and I only realized when I went to deploy and the artifacts weren't there

Page-:
@klutchell I added a warning if tests aren't being used to encourage the use of tests but not enforce it for the moment

Page-:
image
image

klutchell:
LGTM

PR #216, opened 2022-09-19
Add support for a `flowzone-preinstall` npm script

Page-:
This can be used for installing native dependencies that are required for installing npm dependencies

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #218, opened 2022-09-20
Fix header in README

anujdeshpande:
Change-type: patch


anujdeshpande:
@balena-ci I self-certify!

PR #221, opened 2022-09-22
PoC: Add support for a custom test job rather than just step

Page-:
Change-type: minor


Page-:
This looks to be a dead end as you get issues like
```
error parsing called workflow "product-os/flowzone/.github/workflows/flowzone.yml@custom-job": error parsing called workflow "./.github/workflows/custom-test.yml": failed to fetch workflow: workflow was not found.
```
because the referenced workflow must exist even if it will not be called

PR #222, opened 2022-09-22
Add support for a custom job matrices

Page-:
This adds the ability for the custom test/publish jobs to make use of matrices and enhances the ability to prototype matrix-using jobs for inclusion into flowzone directly


Page-:
@balena-ci I self-certify!

PR #228, opened 2022-09-26
Introducing a website type

anujdeshpande:
This should help with migrating projects off Landr.

Change-type: minor

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -57,6 +57,12 @@ on:
       COMPOSE_VARS:
         description: "Optional base64 encoded docker-compose `.env` file for testing Docker images"
         required: false
+      CF_ACCOUNT_ID:
+        description: "Cloudflare account ID"
+        required: false
+      CF_API_TOKEN:
+        description: "Cloudflare API token with limited access for Pages projects"
+        required: false
       CUSTOM_JOB_SECRET_1:
         description: "Optional secret for using with custom jobs"
         required: false
@@ -146,6 +152,11 @@ on:
         type: string
         required: false
         default: '["ubuntu-latest"]'
+      cloudflare_website:
+        description: "Setting this to your existing CF pages project name will generate and deploy a website. Skipped if empty."
+        type: string
+        required: false
+        default: ""
       custom_test_matrix:
         description: "Comma-delimited string of values that will be passed to the custom test action"
         type: string
@@ -268,6 +279,7 @@ jobs:
       python_poetry: ${{ steps.python_poetry.outputs.enabled }}
       python_versions: ${{ steps.python_versions.outputs.json }}
       cargo: ${{ steps.cargo.outputs.enabled }}
+      has_readme: ${{ steps.has_readme.outputs.enabled }}
 
       custom_test: ${{ steps.custom.outputs.test }}
       custom_test_matrix: ${{ steps.custom_test_matrix.outputs.build }}
@@ -553,6 +565,19 @@ jobs:
           if test -f "Cargo.toml"
           then
             echo "found Cargo.toml"
+            echo "enabled=true" >> $GITHUB_OUTPUT
+          else
+            echo "enabled=false" >> $GITHUB_OUTPUT
+          fi
+
+      # Check if we have atleast a README to build a website from
+
+      - name: Check for README for building a website
+        id: has_readme
+        run: |
+          if test -f "README.md"
+          then
+            echo "found README.md"
             echo "enabled=true" >> $GITHUB_OUTPUT
           else
             echo "enabled=false" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


anujdeshpande:
@ab77 I am trying to use the flowzone.yml I edited in this PR/branch in a test repo like [this](https://github.com/balena-io-hardware/hardware-process/blob/test-flowzone/.github/workflows/flowzone.yml) and running into some issues

- we don't edit the `.github/workflows/flowzone.yml` manually, so my PR here edits the `/flowzone.yml`. But GitHub won't treat that as a reusable workflow. 
- Is there a better way to test my changes while they are here on a feature branch?

ab77:
> @ab77 I am trying to use the flowzone.yml I edited in this PR/branch in a test repo like [this](https://github.com/balena-io-hardware/hardware-process/blob/test-flowzone/.github/workflows/flowzone.yml) and running into some issues
> 
> * we don't edit the `.github/workflows/flowzone.yml` manually, so my PR here edits the `/flowzone.yml`. But GitHub won't treat that as a reusable workflow.
> * Is there a better way to test my changes while they are here on a feature branch?
> 
> Edit - I added the file built after `npm run build`, and will remove it later. But still wondering if this is indeed the best way to go about testing a WIP branch in flowzone

@anujdeshpande you just reference your PR workflow using @{{branch}}. If you run `npm i` in your flowzone PR, it will beuild the yaml file correcrtly (e.g. https://github.com/balena-io/etcher/pull/3819)

anujdeshpande:
@ab77 got it - that's what I was doing. I was just thrown off by the fact that I have to commit 2 flowzone.ymls on my PR to be able to test it

anujdeshpande:
This PR now depends on https://github.com/product-os/docusaurus-theme/pull/1

flowzone.yml
```diff
@@ -1398,7 +1422,38 @@ jobs:
...
 
+  website_build:
+    name: Build website using Docusaurus
+    runs-on: ${{fromJSON(inputs.runs_on)}}
+    needs: [project_types,versioned_source]
+    if: |
+      needs.project_types.outputs.cloudflare_website == 'true' &&
+      inputs.cloudflare_website != ''
+    steps:
+      - uses: actions/checkout@v3
```
Page-, line 1437: This should be using
```suggestion
      - *downloadSourceArtifact
      - *extractSourceArtifact
```
so that the versioned source is what is used

flowzone.yml
```diff
@@ -1398,7 +1422,38 @@ jobs:
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - name: Docusaurus Theme
+        run: |
+          git clone https://github.com/product-os/docusaurus-theme
+          cd docusaurus-theme
+          chmod +x docusaurusify.sh
+          # Replace params here with variables
+          ./docusaurusify.sh ${{ github.event.repository.name }} ${{ github.repository_owner }} ${{ github.event.repository.default_branch }} https://${{ inputs.cloudflare_website }}.pages.dev
```
Page-, line 1447: This makes me think it might make more sense for `https://github.com/product-os/docusaurus-theme` to expose a github action that we reuse here like the `balena-io/deploy-to-balena-action` in https://github.com/balena-io/deploy-to-balena-action/blob/master/action.yml since right now there needs to be fairly intimate knowledge of how to run `product-os/docusaurus-theme` built into flowzone vs just calling an action with the right variables

Page- (commented):
You also need to check reviews on
https://github.com/product-os/docusaurus-theme/pull/5
https://github.com/product-os/docusaurus-theme/pull/4
https://github.com/product-os/docusaurus-theme/pull/1

flowzone.yml
anujdeshpande: @Page- good idea - will convert it and update this PR

anujdeshpande: Done!

flowzone.yml
```diff
@@ -571,6 +583,18 @@ jobs:
           cmd: bash -c "echo $INPUT | tr -d '[:space:]'"
           separator: ","
 
+      # Check if we have atleast a README to build a website from
+      - name: Check for README for building a website
+        id: cloudflare_website
+        run: |
+          if test -f "README.md"
```
Page-, line 590: You're going to build a cloudflare page for every single repo we have? I believe they all have a readme (I think balenaCI required a `README.md` even if it was empty). Maybe that makes sense but I'm not sure

anujdeshpande: @Page- it requires at least a readme and most definitely an input called "cloudflare_website" in your repo level flowzone.yml
I am checking for only the readme in this part of the yml - and checking for the input further down.

Most definitely don't intend to build a site for each repo. Only ones that want that and express so through their inputs

anujdeshpande: PS - do you suggest I check for the input at the top instead of at the bottom right before I do the actual building?

Page-: I think renaming this to `has_readme` might make sense and then the `website_build` is `has_readme && cloudflare_website != ''`

anujdeshpande: Sounds good!

anujdeshpande:
@balena-ci I self-ceritfy!

anujdeshpande:
@balena-ci I self-certify!

PR #229, opened 2022-09-26
Don't try to publish to balena if `balena_slugs` is empty

Page-:
Change-type: patch


Page-:
The description of `balena_slugs` already says this will happen but in practice it did not

Page-:
@balena-ci I self-certify!

PR #234, opened 2022-09-27
Use gzip compression for the source artifact

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #235, opened 2022-09-27
Use the `FLOWZONE_TOKEN` not `GITHUB_TOKEN` for publish docs

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #237, opened 2022-09-27
Use gzip compression for the docker image artifact

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #239, opened 2022-09-28
updates dockerhub docs

otaviojacobi:
change-type: patch


otaviojacobi:
Hello @Page- This is a very very minor detail but considering many people are changing to Flowzone now and are using this template for their docker projects I think this bit can get messy in the future if people use the old variable.

If I understood this wrongly and it should still be `DOCKER_REGISTRY_USER/DOCKER_REGISTRY_PASS` please ignore.

Page-:
So `DOCKER_REGISTRY_USER` is what's actually set on all the orgs atm but I agree the names of the secrets in flowzone and the org should match so I'll just duplicate the org secrets with both names and we can start transitioning to standardized naming

Page- (approved):

PR #242, opened 2022-09-28
Ensure finalize events always have a sha

klutchell:
Some events like pushing a tag are valid but will not have a pull_request event body.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #244, opened 2022-09-28
Only allow tag events if versioning is disabled

klutchell:
Manual tagging of the main branch is effectively manual versioning, and would conflict with automatic Flowzone versioning.

This way Flowzone cannot be triggered by previous Flowzone versioned commits.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #247, opened 2022-09-30
flowzone: remove check for default branch on pull request events

alexgg:
This allows to work with other branches that are not the default, specifically ESR branches on meta-balena.

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -137,7 +137,7 @@ jobs:
     timeout-minutes: ${{ inputs.jobs_timeout_minutes }}
     # all jobs depend on this one in some way so add global trigger rules here
     if: |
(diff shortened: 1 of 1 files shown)


Page- (commented):
I'd prefer to have the commit subject (and hence changelog entry) focus on the *why* rather than the *what* - the *what* can be seen by looking at the change itself and the *why* is the bit that is important to note as otherwise it might be lost. ie could you change the commit message to
```
Support pull requests against any branch rather than only the default

Specifically this allows working with ESR branches on meta-balena.

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>
```
or something similar in spirt

alexgg:
@balena-ci I self-certify!

Issue #230, opened 2022-09-26
Add support for multiple buildx bake targets


klutchell:
We are using docker buildx .hcl files to support advanced docker options like platforms, dockerfile, target stage, etc.

However, this still limits us to one bake target when there are valid use cases for producing multiple image variants by specifying a different context, or dockerfile, or build stage.

By supporting a list of bake targets as an input we could build them all. If more than one is provided we could use the bake target as the `prefix` in the docker metadata action.

I'll provide examples of how this could work when I get a chance. The current outstanding use case is open-balena-base which currently builds both systemd and non-systemd variants.

klutchell:
Actually the 'build' target should always be the default and not result in a prefix. Any other target name must inherit build and will use the target name as the prefix.

The tricky bit will be figuring out how to populate the multiple dimensional build matrix for target/platform since each target could have unique platforms.

PR #249, opened 2022-10-03
Support multiple Docker bake targets

klutchell:
This commit replaces the Docker platform build/test matrix
with a single docker-compose test. It also allows multiple bake
targets for publishing image variants with unique tags.

Currently only the linux/amd64 platform image will be tested
with compose tests but any targets defined in the bake file(s)
will be published as defined.

The same cache scope is used for the compose tests as the publish
steps so targets with long build times should need to build once.
Additional cache sources can be added to the bake file(s).

Change-type: minor
Resolves: https://github.com/product-os/flowzone/issues/230
Resolves: https://github.com/product-os/flowzone/issues/253

PRs for testing:
- https://github.com/balena-io-modules/open-balena-base/pull/178
- https://github.com/klutchell/unbound-docker/pull/100

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -53,7 +53,7 @@ on:
         description: "GitHub Actions runner type."
         type: string
         required: false
-        default: "[\"ubuntu-latest\"]"
+        default: '["ubuntu-latest"]'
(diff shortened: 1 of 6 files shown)


klutchell:
@balena-ci I self-certify!

PR #251, opened 2022-10-05
Remove default value for balena_slugs input

klutchell:
It's rare that GH repo slugs align with balenaCloud application slugs so default to an empty list.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #254, opened 2022-10-10
Always run custom action even if workflow is cancelled

ab77:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -225,6 +225,7 @@ jobs:
       custom_finalize: ${{ steps.custom.outputs.finalize }}
       custom_finalize_matrix: ${{ steps.custom_finalize_matrix.outputs.build }}
       custom_clean: ${{ steps.custom.outputs.clean }}
+      custom_always: ${{ steps.custom.outputs.always }}
 
     steps:
       - name: Checkout source
@@ -383,6 +384,10 @@ jobs:
           if [ -d .github/actions/clean ]
           then
             echo "::set-output name=clean::true"
+          fi
+          if [ -d .github/actions/always ]
+          then
+            echo "::set-output name=always::true"
           fi
 
       - name: Convert custom_test_matrix to a JSON array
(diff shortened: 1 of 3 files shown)


ab77:
@klutchell any objections to run custom clean if the workflow is cancelled? The issue with cancelling the entire workflow run due to concurrency setting means that currently none of the cleanup steps in the actual custom actions are executed, leaving behind EC2 instances, etc.

klutchell:
@ab77 Currently custom clean actions are used to remove GH pages preview branches for some repos. But I guess a cancelled run due to concurrency will also have a new run re-create the branch?

We might want to consider a new action for cancelled runs, to keep it separate from closed PRs?

Maybe we need to rename them to opened/merged/cancelled/closed to be more obvious?

ab77:
> Maybe we need to rename them to opened/merged/cancelled/closed to be more obvious?

@klutchell yeah, or maybe we leave the cleanup as is, and add an `always` action, which would be run based on `if: always()` test?

klutchell:
Yeah, like some kind of always post run step is something I've been looking
for but never quite got the criteria or use case right. It sounds like this
is a good case.

.github/workflows/flowzone.yml
```diff
@@ -1334,6 +1334,38 @@ jobs:
...
+        with:
+          name: source-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
+          path: /tmp
+
+      - name: Extract source artifact
+        working-directory: .
+        shell: bash
+        run: tar -xvf /tmp/source.tgz
+
+      - uses: ./.github/actions/always
```
klutchell, line 1363: We need to add checks that this file exists, like we've done for the other custom actions.

When running other repos the example actions do not exist.

.github/workflows/flowzone.yml
```diff
@@ -1334,6 +1334,38 @@ jobs:
...
+    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
+    needs:
+      - event_types
+      - project_types
+      - versioned_source
+      - custom_test
+      - custom_publish
+      - custom_finalize
+      - custom_clean
+    if: always()
```
klutchell, line 1349: I don't think this should run if event_types is skipped or failed?

klutchell (changes requested):

.github/workflows/flowzone.yml
ab77: @klutchell I've just made it dependent on any of the other `test_*` events. I think that should be safe.

klutchell: This job needs to depend on `needs.project_types.outputs.custom_always == 'true'` and I still think we need to depend on `needs.event_types.result == 'success' to make sure we don't run if the event type cannot be parsed or is skipped.

klutchell (changes requested):

.github/workflows/flowzone.yml
klutchell: The `custom_always` job should not run if `actions/custom/action.yml` is not found, it will fail. Like I said, try this on a repo where that file does not exist and you'll see what I mean.

ab77: so basically changing from `always()` to this, right?

```
...
  custom_always:
    name: Always custom
    runs-on: ${{ fromJSON(inputs.runs_on) }}
    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
    needs:
      - event_types
      - custom_test
      - custom_publish
      - custom_finalize
      - custom_clean
    if: needs.project_types.outputs.custom_always == 'true'

    steps:
...
```

klutchell:
```suggestion
    if: !failure() &&
      needs.event_types.result == 'success' &&
      needs.project_types.outputs.custom_always == 'true'
```

This would always run unless a previous job failed. If you still want to run in the case of failures change `!failure()` back to `always()` but keep the other two checks.

PR #256, opened 2022-10-11
Accept BALENA_TOKEN in addition to BALENA_API_KEY

cywang117:
This allows `secrets: inherit` to be used without issue, as the deploy-to-balena action requires BALENA_TOKEN but Flowzone only passed BALENA_API_KEY.

Change-type: patch
Closes: https://github.com/balena-io/deploy-to-balena-action/issues/213
Signed-off-by: Christina Ying Wang <christina@balena.io>

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -33,6 +33,9 @@ on:
       BALENA_API_KEY:
         description: "API key for pushing releases to balena applications"
         required: false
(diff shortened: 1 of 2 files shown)


Page- (changes requested):
I think downstream should only use `BALENA_API_KEY`, it's only legacy code that predates flowzone that uses other names for the various flowzone secrets and needs to update them to the standardized interface

PR #258, opened 2022-10-12
Debug npm pack

ab77:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -656,7 +656,11 @@ jobs:
         run: npm test
 
       - name: Run pack
-        run: mkdir /tmp/npm-pack && npm pack --pack-destination=/tmp/npm-pack
+        run: |
+          mkdir /tmp/npm-pack && npm pack --pack-destination=/tmp/npm-pack
+
+          # FIXME: workaround when npm pack dumps tarball into the current directory
+          find . -name '*.tgz' -exec mv {} /tmp/npm-pack \;
 
       - name: Upload artifact
         uses: actions/upload-artifact@v3


ab77:
@klutchell this is the best I was able to come up with. Seems like at least in versionbot's instance, npm pack puts the tgz in the current working directory. Could be something weird with permissions/etc., but the error is silent and would be tricky to debug further. There is a closed issue in GH node/npm archive about "inconsistent npm pack output", but no resolution.

Page-:
I kind of suspect this is due to using an old npm version on node 12, checking the npm 6 docs I don't see the `--pack-destination`: https://docs.npmjs.com/cli/v6/commands/npm-pack

ab77:
> I kind of suspect this is due to using an old npm version on node 12, checking the npm 6 docs I don't see the `--pack-destination`: https://docs.npmjs.com/cli/v6/commands/npm-pack

Yeah, it's definitely a node12 thing, since I was able to replicate locally. Should we keep this hack or take it out and assume node12 builds will fail? @Page- @klutchell

.github/workflows/flowzone.yml
```diff
@@ -1351,10 +1355,7 @@ jobs:
       - custom_publish
       - custom_finalize
       - custom_clean
-    if: |
-      !failure() &&
-      needs.event_types.result == 'success' &&
-      needs.project_types.outputs.custom_always == 'true'
```
klutchell, line 1357: Was this change committed by mistake?

klutchell (changes requested):

.github/workflows/flowzone.yml
ab77: @klutchell this is intentional, the current check still skips the always step when the pipeline is cancelled/fails.

I think it should be a very simple check instead:

```
if: needs.project_types.outputs.custom_always == 'true'
```

.. with the caveat that the always custom action should return `true` even if the cleanup fails.

ab77: basically, if the user has `always/action.yml`, we *always* need to run it, since it's important (otherwise expensive assets maybe left running).

klutchell: What if a previous step fails? Do we still want/need to run the custom step?

At the very least this change should be a separate commit, if not a separate PR. It's very confusing otherwise.

ab77: > What if a previous step fails? Do we still want/need to run the custom step?

that's the whole point of `always()` - if something fails, we still want to ensure the clen-up steps are done. This would be ideally handled in the composite action, but the GH approach is to kill them without respecting their own `always()` steps.

ab77: > At the very least this change should be a separate commit, if not a separate PR. It's very confusing otherwise.

@klutchell agree, this is a shortcut on my part, I'll move it

PR #259, opened 2022-10-12
Update set-output commands to use the new out environment file

Page-:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Change-type: patch


Page-:
@balena-ci I self-certify!

.github/workflows/flowzone.yml
```diff
@@ -1524,4 +1524,7 @@ jobs:
             exit 1
           fi
 
-          echo ::set-output name=result::"${result}"
+          DELIMITER=$(echo $RANDOM | md5sum | head -c 32)
```
Page-, line 1527: The random delimiter is to avoid issues if the delimiter happens to be contained in the output

PR #263, opened 2022-10-13
Always run `custom_always` step

ab77:
* if action.yml is present regadless of status of other jobs

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -1356,8 +1356,7 @@ jobs:
       - custom_finalize
       - custom_clean
     if: |
-      !failure() &&
-      needs.event_types.result == 'success' &&
+      always() &&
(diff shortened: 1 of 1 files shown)


.github/workflows/flowzone.yml
```diff
@@ -1351,10 +1351,7 @@ jobs:
       - custom_publish
       - custom_finalize
       - custom_clean
-    if: |
-      !failure() &&
-      needs.event_types.result == 'success' &&
-      needs.project_types.outputs.custom_always == 'true'
+    if: needs.project_types.outputs.custom_always == 'true'
```
klutchell, line 1354: You still need the `always()` here if removing the `!failure()`, but then it should be good.

klutchell (changes requested):

klutchell (approved):

PR #266, opened 2022-10-14
flowzone: Support build metadata in semvers for device repositories

alexgg:
Device repositories use build meta-data so let's not remove it from the version used in versioning.

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>


.github/workflows/flowzone.yml
```diff
@@ -530,7 +530,12 @@ jobs:
           [ -f package.json ] && versions+=($(jq -r .version package.json))
           [ -f balena.yml ] && versions+=($(yq e '.version' balena.yml))
           [ -f VERSION ] && versions+=($(head -n1 VERSION))
-          semver="$(npx -q -y -- semver "${versions[@]}" | tail -n1)"
+          npm install -g semver > /dev/null 2>&1
+          semver="$(node <<- EOF
+          const semverParse = require('semver/functions/parse')
+          console.log(semverParse("${versions[0]}")['raw'])
+          EOF
+          )"
```
Page-, line 538: What's the difference between this and just doing
```suggestion
          semver="${versions[0]}"
```
?

alexgg: My thinking was that there is an implicit semver validity check in using the semver module @Page- If that's not relevant why would it have been not done like that initially?

alexgg: Redid the PR to the simpler version and amended the commit with reasoning.

alexgg:
More review comments in [zulip](https://balena.zulipchat.com/#narrow/stream/345889-loop.2Fbalena-os/topic/Switching.20balenaCI.20for.20flowzone.20and.20all.20device.20repositories) (internal)

alexgg:
@balena-ci I self-certify!

PR #267, opened 2022-10-14
Only check in the root of the working dir for npm pack files

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #269, opened 2022-10-14
Only look for an npm pack in the working dir for node 6.x

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #270, opened 2022-10-14
Combine docker tests and publish steps

klutchell:
This will allow testing of the same image that is published
by using `image: localhost:5000/sut` in a compose file.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #273, opened 2022-10-17
Revert merging of docker test/publish steps

Page-:
These changes are causing problems with some repos, likely related ti testing docker images without publishing them

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -455,7 +455,7 @@ jobs:
         shell: bash --noprofile --norc -eo pipefail -x {0}
 
     outputs:
-      new_tag: ${{ steps.new_version.outputs.tag || steps.git_describe.outputs.tag }}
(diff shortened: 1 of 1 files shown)


Page-:
https://github.com/product-os/jellyfish-plugin-feedback/actions/runs/3257660196/jobs/5349115154

pipex:
I was about to bring this issue up https://github.com/balena-os/balena-supervisor/actions/runs/3265448166/jobs/5368938097 

I suspect because of the changes the `join()` on `docker_images` no longer returns a zero length string https://github.com/product-os/flowzone/blob/f160594a21a1fec8e287b5c2d1e3ae0a9a27a53b/.github/workflows/flowzone.yml#L918

klutchell:
Thanks for reverting this @Page- and @pipex , I'll take another shot at these changes when I get the chance and re-test on the supervisor before merging.

PR #274, opened 2022-10-17
Update readme to better explain the `protect_branch: false` intent

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #276, opened 2022-10-17
Combine docker test and publish jobs

klutchell:
This will allow testing of the same image that is published by using image: `localhost:5000/sut` in a compose file.

It also ensures only one build and improved caching.

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -849,113 +849,17 @@ jobs:
     runs-on: ${{ fromJSON(inputs.runs_on) }}
     timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
     needs:
-      - event_types
-      - project_types
-      - versioned_source
-    if: |
-      needs.event_types.outputs.do_draft == 'true' &&
-      needs.project_types.outputs.docker_compose_test != ''
-    defaults:
-      run:
-        working-directory: ${{ inputs.working_directory }}
-        shell: bash --noprofile --norc -eo pipefail -x {0}
-    strategy:
-      fail-fast: false
-      matrix:
-        target:
-          - default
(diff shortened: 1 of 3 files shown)


.github/workflows/flowzone.yml
```diff
@@ -991,40 +897,71 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: |
-            ${{ env.DOCKER_IMAGES }}
+            ${{ env.DOCKER_IMAGES || env.SUT_IMAGE }}
```
klutchell, line 900: The previous merge was missing this fallback for when the images list is empty. We still want to build and push to the local docker registry for testing.

.github/workflows/flowzone.yml
```diff
@@ -990,41 +896,76 @@ jobs:
...
 
-      # this override file will add additional cache sources and inherit the docker-metadata-action
+      # this override file will add additional cache sources and inherit the docker-metadata-action target
       # https://docs.docker.com/build/customize/bake/file-definition/#json-definition
       - name: Create bake override
         run: |
-          jq -n '{target:{default:{}}}' > ${BAKE_EMPTY}
+          jq -n '{target:{"${{ matrix.target }}":{}}}' > ${BAKE_EMPTY}
           docker buildx bake --print ${{ matrix.target }} \
             -f ${{ join(fromJSON(needs.project_types.outputs.docker_bake),' -f ') || env.BAKE_EMPTY }} \
```
klutchell, line 916: @Page- Plain dockerfile support should be restored here. As long as there are docker images provided as an input, it should build and test them.

Though do you mean we should build even if not publishing or testing? Seems like a waste for repos that may be intended for balena push only?

klutchell:
@balena-ci I self-certify!

PR #278, opened 2022-10-18
Custom OS matrix

ab77:
Change-type: patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,11 +319,19 @@ They can also be found at the top of [flowzone.yml](./.github/workflows/flowzone
 
 #### `runs_on`
 
-GitHub actions runner (e.g. `[\"self-hosted\",\"linux\",\"x64\"]` for self-hosted runners.)
+GitHub actions runner for core jobs (e.g. `["self-hosted"]` for self-hosted runners.)
 
 Type: _string_
 
-Default: `[\"ubuntu-latest\"]`
+Default: `["ubuntu-latest"]`
+
+#### `tests_run_on`
+
+GitHub Actions runner type for tests (e.g. multiple OS platforms, multiple Linux versions, etc.)
+
+Type: _string_
+
+Default: `["ubuntu-latest"]`
 
 #### `jobs_timeout_minutes`
 
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -4,13 +4,14 @@
     uses: actions/download-artifact@v3
     with:
       name: source-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
-      path: /tmp
+      path: ${{ runner.temp }}
 
   - &extractSourceArtifact
     name: Extract source artifact
+    # tar: Cannot connect to D: resolve failed
+    shell: pwsh
     working-directory: .
-    shell: bash
-    run: tar -xvf /tmp/source.tgz
+    run: tar -xvf ${{ runner.temp }}/source.tgz
 
 name: Flowzone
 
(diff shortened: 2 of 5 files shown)


flowzone.yml
```diff
@@ -4,13 +4,20 @@
...
+    if: runner.os != 'Windows'
     working-directory: .
-    shell: bash
-    run: tar -xvf /tmp/source.tgz
+    run: tar -xvf ${{ runner.temp }}/source.tgz
+
+  - &extractSourceArtifactWindows
+    name: Extract source artifact
+    if: runner.os == 'Windows'
+    shell: powershell
```
Page-, line 18: Is explicitly choosing powershell required to run on windows?

flowzone.yml
```diff
@@ -4,13 +4,20 @@
...
+    working-directory: .
+    run: tar -xvf ${{ runner.temp }}\source.tgz
```
Page-, line 20: It should also with fine with `/` on windows, so this should be able to match the linux version
```suggestion
    run: tar -xvf ${{ runner.temp }}/source.tgz
```

README.md
```diff
@@ -397,6 +397,14 @@ Type: _string_
 
 Default: `''`
 
+#### `custom_runs_on`
+
+GitHub Actions runner type for custom jobs.
```
Page-, line 402: Why do we need two separate `runs_on` entries?

flowzone.yml
ab77: Yeah, there was a reason for it. I think it was to do with tar.

ab77: @Page- I am hesitant to go through retesting it all from the top now having spent all that time tuning the Etcher build... I am fairly certain I had to split these jobs or a reason.

flowzone.yml
Page-: It works fine in powershell (6+7 at least), although it can cause issues in cmd - `/` is used for args so eg `dir ./foo` would give `Invalid switch - "foo".` but `dir "./foo"` works

README.md
Page-: Presumably that's only a problem for a handful that we want to be run as singletons, eg `Event types`/`Versioned Source`/`Protect Branch` but for others we may want to do multi-arch like `Test npm` in order to test on multiple OS versions. I'm thinking it may make more sense to have the distinction of `runs_on` vs `tests_runs_on`?

flowzone.yml
Page-: The main reason is it'd let us drop a bunch of code if we can use the same block for both, and it looks like that should be fine since the default shell for windows is powershell anyway: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
Although it'd also be possible to specify `bash` and that should still work fine on windows, or `pwsh` and that should also work fine on linux

README.md
ab77: @Page- do you want me to rename `custom_runs_on` to `tests_runs_on`?

flowzone.yml
ab77: @Page- here's the issue: https://github.com/balena-io/etcher/actions/runs/3386762422/jobs/5631091011#step:3:10

flowzone.yml
ab77: tar doesn't work when running on Windows and with bash as the shell.

Page-: Try `shell: pwsh`

ab77:
@Page- any more comments?

ab77:
@balena-ci I self-certify!

PR #279, opened 2022-10-18
Fix for empty COMPOSE_VARS

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #281, opened 2022-10-18
Add Python Poetry test to flowzone.

gantonayde:
Add python_versions to inputs.
Add Python info to README.
Add Python poetry test project to tests.

Change-type: minor
Signed-off-by: Genadi Naydenov <genadi@balena.io>


pyproject.toml
```diff
@@ -0,0 +1 @@
+tests/pyproject.toml
```
Page-, line 1: This file should be unnecessary

.github/workflows/flowzone.yml
```diff
@@ -84,6 +84,11 @@ on:
         type: string
         required: false
         default: ""
+      python_versions:
```
Page-, line 87: Is there a way to specify supported python versions native to python itself, maybe in `pyproject.toml` somehow? If so we could fetch the versions to run against from that file rather than having a separate CI specifier

gantonayde: This was my initial attempt but there were two issues which I could not solve. The first one is that I am not sure how to convert multiple inequalities, i.e. `>=3.7,<3.10,!=3.8` into a list for the matrix `[3.7, 3.9]` as poetry often requires the Python versions to be specified with inequalities (packages can become unhappy if a caret is used for the python version in the toml file, here is an [example](https://github.com/python-poetry/poetry/blob/master/docs/faq.md#why-is-poetry-telling-me-that-the-current-projects-python-requirement-is-not-compatible-with-one-or-more-packages-python-requirements)). Just a note that the Go version of `yq` that comes preinstalled on the runner does not support `.toml` files and something like `dasel` needs to be downloaded if we want to parse the .toml properly.

The second one is that poetry does not have something nice like npm check-engine to return a status.

Now when I think about it again the closest thing is probably `poetry install –dry-run` which will return
```
The currently activated Python version 3.8.13 is not supported by the project (~3.9).
Trying to find and use a compatible version.

  NoCompatiblePythonVersionFound
```
I guess checking for `NoCompatiblePythonVersionFound` can be used. I will try with this one to repeat the Node version selection steps.

.github/workflows/flowzone.yml
```diff
@@ -387,6 +389,111 @@ jobs:
...
+      - name: Check compatibility
+        if: steps.python_poetry.outputs.enabled == 'true'
+        run: |
+          error_check=`(poetry install --dry-run --only-root 2>&1 || true)`
+          if ! grep -q "Please change python\|Poetry was unable to find a compatible version" <<< $error_check
+          then
+            echo "PYTHON_VERSIONS=$(echo "${PYTHON_VERSIONS}" | jq -c '. + ["3.7"]')" >> $GITHUB_ENV
+          else
+            echo "Python 3.7 does not meet project requirements."
+          fi
```
gantonayde, line 432: This bit works here but fails for the [bottom-up](https://github.com/company-os/bottom-up-model-python/actions/runs/3282958824/jobs/5407059428#step:23:31) project. I need to investigate.

gantonayde: Okay, so poetry was not changing the `python env` accordingly to the Python setup during the checks. 
Apparently, `poetry env use` also triggers a check whether python meets the project requirements which is a lot quicker as other dependencies do not need to be resolved. Hence using `poetry env use` instead of `poetry install` to deduce if the current python version is fine saves a lot of time, i.e. for the bottom-up project the compatibility check went from ~2mins to 9 secs.

.github/workflows/flowzone.yml
```diff
@@ -1199,6 +1306,77 @@ jobs:
           versionbot: false # disable the included versionbot branch checkout
           source: ${{ inputs.working_directory }}
 
+  ###################################################
+  ## Python
+  ###################################################
+
+  python_test:
```
Page-, line 1313: You should make sure any steps that need to depend on `python_test`, at the very least the `protect_branch` job should depend on it

gantonayde:
Summary:

- Add Python Poetry test to flowzone.
- Deduce suitable Python versions (=>3.7,<3.11) for the test.
- Add Python (Poetry) test project to flowzone tests.
- Update the README.md accordingly.

.github/workflows/flowzone.yml
```diff
@@ -387,6 +389,111 @@ jobs:
...
+              echo "PYTHON_VERSIONS=[]" >> $GITHUB_ENV
+            else
+              echo "Poetry not used"
+              echo "enabled=false" >> $GITHUB_OUTPUT
+            fi
+          else
+            echo "enabled=false" >> $GITHUB_OUTPUT
+          fi
+
+      # Check which Python 3.7+ versions meet the Poetry project requirements
```
pranasziaukas, line 411: Would be awesome to have a loop here instead of hardcoding the versions. But I suppose that is not really feasible in this situation?

pranasziaukas (commented):
Nice work Genadi, and looks to be fully in line with the company-wide adoption of Flowzone 👍

.github/workflows/flowzone.yml
```diff
@@ -1199,6 +1306,77 @@ jobs:
...
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Poetry
+        run: |
+          pipx install poetry
```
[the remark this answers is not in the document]
gantonayde: Pipx is installed by default on the runner and adds poetry in a separate environment that is globally accessible, whereas with curl you also need to include poetry to the path in some cases. Plus it is easier to read it like that.

.github/workflows/flowzone.yml
gantonayde: I think looping is done via matrices and maybe there is a way but I followed the Node checks template to keep everything consistent.

.github/workflows/flowzone.yml
```diff
@@ -387,6 +389,111 @@ jobs:
...
+          error_check=`(poetry env use 3.10 2>&1 || true)`
+          if ! grep -q "Please choose a compatible version" <<< $error_check
+          then
+            echo "PYTHON_VERSIONS=$(echo "${PYTHON_VERSIONS}" | jq -c '. + ["3.10"]')" >> $GITHUB_ENV
+          else
+            echo "Python 3.10 does not meet project requirements."
+          fi
+
+      # default to the latest version (^3.7) on the runner 
+      # if none were matched by the checks above
```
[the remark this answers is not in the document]
gantonayde: Not really, it does not check what [versions](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#python) the runner provides. I guess this can be improved along with the comment above and applied to Node and npm tests as well.

.github/workflows/flowzone.yml
Page-: A matrix happens for the entire job @pranasziaukas / @gantonayde so it doesn't work so well for a job like this where we only want to matrix specific steps (if I'm wrong please tell me/ping on zulip because it'd be awesome to have per-step matrices!)

gantonayde: Alas I could not make it work, as @Page- explained and as per their docs you cannot generate steps via loops at the moment, which is a bit sad.

Issue #280, opened 2022-10-18
Docker bake breaks when the context directory referenced by a docker-compose file is not in the root directory


cmfcruz:
The docker bake breaks when the context directory referenced by a docker-compose file is not in the root directory of the repository. 

- [Sample job run that failed](https://github.com/balena-io/environment-staging/actions/runs/3273114499/jobs/5384950534)
- [docker-compose.yml file in one of the subdirectories](https://github.com/balena-io/environment-staging/blob/master/balenaCloud/balena-engine/balena-engine-amd64/docker-compose.yml)
- [docker-compose.yml file in root directory](https://github.com/balena-io/environment-staging/blob/master/docker-compose.yml)

cmfcruz:
There is inconsistent behavior in docker buildx bake where the first docker-compose yaml determines the working directory for the path resolution: https://github.com/docker/buildx/issues/197

PR #282, opened 2022-10-18
Only look for docker-bake/compose files in the root of the working dir

Page-:
This avoids issues with ones in the nested dirs that are not expected to be used by default

Change-type: patch


Page-:
Fixes #280

PR #283, opened 2022-10-18
Plain dockerfile support

Page-:
Change-type: patch

### DIFF
--- a/.github/workflows/test-plain-dockerfile.yml
+++ b/.github/workflows/test-plain-dockerfile.yml
@@ -1,0 +1,20 @@
+name: Test plain dockerfile
+
+on:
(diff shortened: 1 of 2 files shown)


klutchell:
What are you expecting to have happen when only a Dockerfile is present? Unless we are publishing or testing something, I would think we would want to avoid building Docker. Especially for repos like [this](https://github.com/balenalabs/balena-sense) where a Docker build is expected to fail since it's meant to be used with balena push for a specific architecture.

I think without compose tests, or public registry names, we shouldn't do anything Docker related. Maybe there is a case to be made that if bake files exist we do a Docker build at least, but not if only a Dockerfile is found.

PR #285, opened 2022-10-18
Only check for version tags from the current branch

Page-:
This avoids issues where a PR is 1 version behind master (eg another PR was recently merged) and then the updated version for the PR branch will match the master version and give a false positive about "Failed to detect any versioned files"

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #287, opened 2022-10-19
Fail if generating CHANGELOG.yml fails since the bug was fixed

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #288, opened 2022-10-19
Use a build step for flowzone.yml to support yaml features like anchors

Page-:
This will happen automatically on commit as long as you have run `npm install` to set up the git hooks/dependencies

Change-type: patch


Page-:
In the aggregate PR github doesn't handle the rename file + add one with the same name as a previous one very nicely but if you look at it commit-by-commit or at the history/blame for `/flowzone.yml` then you'll see the history is tracked correctly (and means that rebases should work nicely)

Page-:
@balena-ci I self-certify!

PR #289, opened 2022-10-19
Support fetching submodule commits not in the repository HEAD

alexgg:
As [discussed](https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/balena-phytec.20checkout) we either change the default `fetch-depth` in project_types to `0` or we make just that configurable, which is this patch.

I'd go for the simplest approach, but that probably affects the overall speed of flowzone so this PR is a compromise with a worst UX (we use `fetch-depth` in two jobs but the global input only configures one.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -104,6 +104,11 @@ on:
         type: string
         required: false
         default: "Flowzone"
+      checkout_fetch_depth:
+        description: "Configures the depth of the actions/checkout git fetch."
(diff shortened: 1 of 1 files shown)


alexgg:
@balena-ci rebase

alexgg:
@balena-ci rebase

alexgg:
@balena-ci I self-certify!

codewithcheese:
LGTM :grin:

PR #292, opened 2022-10-20
Only login to registry if the NPM_TOKEN is present

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #294, opened 2022-10-21
Fix adding protections to a branch that has no existing protections

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #298, opened 2022-10-25
Fix docker pipeline when only `docker-compose.test.yml` is provided

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #300, opened 2022-10-25
Don't publish to npmjs if already published

dsevastianov:
If published package has been downloaded,
npm doesn't allow to unpublish it.

Assuming that the same commit produces the same package,
publishing can be skipped altogether

Prompted by: https://balena.zulipchat.com/#narrow/stream/345890-loop.2Fbalena-io/topic/Test.20docker.20on.20open-balena-api
Change-type: patch


dsevastianov:
@balena-ci I self-certify!

PR #305, opened 2022-10-31
Provide the git describe tag if versioning is disabled

klutchell:
This allows tagged releases for repos where versionist is disabled.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #311, opened 2022-11-12
Remove docker-compose.yml from parsed bake files

klutchell:
Since compose yaml files are often used for balena deployments, and optionally for docker compose tests, let's avoid overloading the file further and ignore it when looking for bake files.

This should avoid a confusing mess of services to build when both compose files and bake files are present in a repo.

Bake files used for docker build and deploy.
Compose files used for docker compose tests and balena deploy.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #312, opened 2022-11-12
Remove compose test file from bake override

klutchell:
This was only being used for image pre-build and will cause issues if it depends on resources in docker-compose.yml.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #314, opened 2022-11-14
Add Rust support

pipex:
Change-type: minor


flowzone.yml
```diff
@@ -1398,6 +1428,136 @@ jobs:
...
+    steps:
+      - name: Download source artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: source-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
+          path: ${{ runner.temp }}
+      - name: Extract source artifact
+        shell: bash
+        working-directory: .
+        run: tar -xvf ${{ runner.temp }}/source.tgz
```
Page-, line 1463:
```suggestion
      - *downloadSourceArtifact
      - *extractSourceArtifact
```

flowzone.yml
```diff
@@ -1398,6 +1428,136 @@ jobs:
...
```
Page-, line 1518:
```suggestion
      - *downloadSourceArtifact
      - *extractSourceArtifact
```

flowzone.yml
```diff
@@ -1398,6 +1433,233 @@ jobs:
...
+      - name: Compress
+        run: |
+          tar -czvf ${{ needs.cargo_test.outputs.package }}-${{ matrix.target }}-v${{ needs.cargo_test.outputs.version }}.tar.gz -C target/${{ matrix.target }}/release ${{ needs.cargo_test.outputs.package }}
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: gh-release-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
+          path: ${{ needs.cargo_test.outputs.package }}-${{ matrix.target }}-v${{ needs.cargo_test.outputs.version }}.tar.gz
+          retention-days: 90
```
ab77, line 1575: I doubt full 90 days is needed here, I would just set this to minimum of 1 day.

flowzone.yml
```diff
@@ -1398,6 +1433,233 @@ jobs:
...
+
+      - name: Publish artifacts
+        uses: softprops/action-gh-release@v1
+        with:
+          name: ${{ github.event.pull_request.head.ref }}
+          tag_name: ${{ github.event.pull_request.head.ref }}
+          draft: true
+          prerelease: true
+          # Publish anything added to with the gh-release name
+          files: ${{ runner.temp }}/*
```
ab77, line 1621: During my testing, this action doesn't re-create a release if one already exists. So if the release has assets from previous PR runs, each subsequent run will add to these (not replace). Which is why I ended up having to delete previous draft releases manually.

pipex: Oh, good catch, I'll add that

pipex: @ab77 what about the artifacts? Do you know if artifacts uploaded to action context with the same name persist across PR runs?

pipex: Nevermind, I just saw that it doesn't

flowzone.yml
```diff
@@ -1398,6 +1437,286 @@ jobs:
...
+        working-directory: .
+        shell: bash --noprofile --norc -eo pipefail -x {0}
+
+    steps:
+      - *downloadSourceArtifact
+      - *extractSourceArtifact
+
+      - name: Delete draft release
+        run: |
+          set -ea
```
ab77, line 1468: @pipex `set -ea` can go, it's not necessary at all for this step.

flowzone.yml
```diff
@@ -1398,6 +1437,286 @@ jobs:
...
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+          components: rustfmt, clippy
+
+      - name: Check formatting
+        run: cargo fmt --check
+
+      - name: Lint with clippy
+        run: cargo clippy --all-features
```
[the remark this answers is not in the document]
pipex: done

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
Page-, line 1: What is this for?

.github/workflows/tests.yml
```diff
@@ -18,3 +18,6 @@ jobs:
         ghcr.io/product-os/flowzone
       balena_slugs: |
         product_os/flowzone
+      cargo_targets: |
+        x86_64-unknown-linux-gnu,
+        aarch64-unknown-linux-gnu
```
Page-, line 23: We probably also want armv7 since most of our devices use that and I think we'd want our rust builds to work on most devices we support?

pipex: Sure, I can add it, although this just tests that flowzone can build for multiple architectures

.prettierrc
pipex: So editors integrated with prettier know what are the conventions on this repo. We use single quote as convention everywhere else so in my case prettier was changing the whole file on save.

.github/workflows/tests.yml
pipex: Sure, I can add it

pipex:
@balena-ci I self certify!

PR #315, opened 2022-11-16
Test Flowzone without custom bake targets

klutchell:
The common use case will be just dockerfiles and compose files
and a default target so it's better to cover those scenarios.

Multiple bake targets is a rare use case and can be tested separately.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #316, opened 2022-11-17
Support `{{ template-file.yml }}` template variable to reduce LOC

20k-ultra:
could adapt to support `{{ env(MY-ENV-VAR) }}` too if interested.

```
const envTemplate = (p = "env\(([a-zA-Z-])*\)") => {
  return templateVariable(p);
};
```

Btw, no flowzone file change will be committed since the generated file from the build script is the same.


build.js
```diff
@@ -1,21 +1,41 @@
-const fs = require('fs');
-const yaml = require('yaml');
+const fs = require("fs");
+const yaml = require("yaml");
+
+let flowzoneSource = fs.readFileSync("./flowzone.yml", "utf8");
+
+// Pattern to find template variables (sub lines with file)
+const templateVariable = (p) => {
+  return new RegExp(`# {{ *${p} *}}`, "g");
```
20k-ultra, line 8: Uses a comment so that yaml lsp does not freak out from this template variable

20k-ultra:
working on the tabbing being respected.

build.js
```diff
@@ -1,21 +1,62 @@
...
+const assert = require("assert");
+const fs = require("fs");
+const yaml = require("yaml");
+
+let flowzoneSource = fs.readFileSync("./flowzone.yml", "utf8");
+
+// Pattern to find template variables (sub lines with file)
+const templateVariable = (p) => {
+  // Tab or spaced comment with curly braces enclosing a variable
+  return new RegExp(`[ \t]*# {{ *${p} *}}`, "g");
```
20k-ultra, line 10: Uses a comment so YAML lsps do not freak out.

build.js
```diff
@@ -1,21 +1,62 @@
...
+      return tabs(numOfTabs) + l;
+    })
+    .join("\n");
+  function tabs(n, tabReturn = "") {
+    if (n !== 0) {
+      return tabs(n - 1, tabReturn + " ");
+    }
+    return tabReturn + " ";
+  }
+}
```
20k-ultra, line 43: would have much rather the regex replace to not remove the tabs/spaces. In a regex tester I was able to sub without removing them so it's a JS replace implementation that I cannot figure out how tell not to do replace that way.

build.js
```diff
@@ -1,21 +1,62 @@
...
+    const templateValue = fs.readFileSync(m[1], "utf8");
+    const sub =
+      spacing.length > 0 // Adding template tabs/spaces to template source for proper indenting
+        ? tabText(templateValue, spacing.length)
+        : templateValue;
+    flowzoneSource = flowzoneSource.replaceAll(fileTemplate(m[1]), sub);
+  }
+}
+
+function tabText(text, numOfTabs) {
```
20k-ultra, line 30: maybe confusing name since yaml does not allow tabs.

Page-: yeah, the terminology you want is `indentText` where indentation can be either tabs or space in general (although yaml enforces spaces)

20k-ultra:
PR should be implemented with https://eemeli.org/yaml/#writing-custom-tags

PR #317, opened 2022-11-17
Fix finalization failures on npm/rust/github

pipex:
This fixes a few bugs in those language jobs
that were preventing the post-merge run from terminating successfully.

- `github_finalize` was missing a step to download the sources
- `cargo_publish` was wrongly failing if `CARGO_REGISTRY_TOKEN` was
  empty
- `npm_finalize` was failing because the test package publishing was
  conflicting with the flowzone package publish.

This commit also adds a new github clean job, to remove any draft
releases left behind after a PR is closed without merging.

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1441,8 +1441,8 @@ jobs:
   # GitHub
   ###################################################
 
-  github_clean:
-    name: Clean draft GitHub release
+  github_prepare:
+    name: Clean previous GitHub draft release
     runs-on: ${{ fromJSON(inputs.runs_on) }}
     timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
     needs:
@@ -1528,6 +1528,9 @@ jobs:
         shell: bash --noprofile --norc -eo pipefail -x {0}
 
     steps:
+      - *downloadSourceArtifact
+      - *extractSourceArtifact
+
       # https://docs.github.com/en/rest/releases
       - name: Finalize GitHub release (if any)
         if: needs.versioned_source.outputs.version != ''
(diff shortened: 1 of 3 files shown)


pipex:
github cleanup step needs reworking

flowzone.yml
```diff
@@ -1554,6 +1557,34 @@ jobs:
...
+  github_clean:
+    name: Clean unused GitHub release
+    runs-on: ${{ fromJSON(inputs.runs_on) }}
+    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
+    needs:
+      - project_types
+      - versioned_source
+      - npm_publish
+    if: |
+      !failure() && !cancelled() &&
```
pipex, line 1569: Is this line needed ?

flowzone.yml
```diff
@@ -1699,25 +1730,17 @@ jobs:
...
-      - name: Download source artifact from PR
-        uses: dawidd6/action-download-artifact@v2
-        with:
-          github_token: ${{ secrets.FLOWZONE_TOKEN }}
-          commit: ${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
-          path: ${{ runner.temp }}
-          workflow_conclusion: success
-          name: source-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
-
+      - *downloadSourceArtifact
```
Page-, line 1733: I think `downloadSourceArtifact` only works within the context of the current action whereas this needs to download an artifact from another action?

pipex:
@balena-ci I self-certify!

flowzone.yml
pipex: While `*downloadSourceArtifact` works here, the history will contain the merge commit and the versioning, commit, which means the release notes are not that nice, see https://github.com/balena-os/fatrw/releases/tag/v0.2.14 for an example. I'll revert to using the PR source in a new PR

PR #318, opened 2022-11-18
Fix error in github_finalize if no release exists

pipex:
The script now looks for a release before continuing with the finalization steps.

Change-type: patch


pipex:
@balena-ci I self certify!

PR #319, opened 2022-11-18
Ensure github prepare job is run before gh publish

pipex:
Change-type: patch


pipex:
@balena-ci I self certify!

PR #320, opened 2022-11-18
Do not include the merge commit in the release notes

pipex:
During merge, we want to use the PR source for creating the release notes instead of the latest source since that will include the merge commit and it will pollute the notes.

See https://github.com/balena-os/fatrw/releases/tag/v0.2.14 for an example

Change-type: patch


pipex:
@balena-ci I self certify!

PR #321, opened 2022-11-21
Allow external contributors

ab77:


klutchell:
Does `inherit` actually work for secrets with external contributors? It would appear not?

ab77:
> Does `inherit` actually work for secrets with external contributors? It would appear not?

It doesn't even trigger the workflow run, so I am not sure if it's secrets or something else; I've opened a [case](https://support.github.com/ticket/personal/0/1887802).

klutchell:
Look at the actions tab, you can see that the runs failed to start because
the FLOWZONE_TOKEN is not set. It just doesn't get reflected on the PR.

ab77:
> Look at the actions tab, you can see that the runs failed to start because the FLOWZONE_TOKEN is not set. It just doesn't get reflected on the PR.

All the runs with that message are `on:pull_request` @klutchell, the runs `on:pull_request_target` don't even start.

PR #330, opened 2022-11-23
 Use latest source when creating github release notes

pipex:
The `github_finalize` job was trying to use the original source of the PR
for creating the release notes. This required the source artifact to
exist at the moment of merge which is not always true as merge can
happen at any point after creating the PR. This changes the job to use
the latest source but exclude the last two commits (merge and version)
from the release notes.

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1518,9 +1518,11 @@ jobs:
           set -ea
 
           artifact_count=0
-          [ -d "${{ runner.temp }}/gh-release-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}" ] && \
-              artifact_count=$(ls "${{ runner.temp }}/gh-release-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}" | wc -l | sed 's/^ *//;s/ *$//')
+          [ -d "$GH_ARTIFACTS" ] && \
+              artifact_count=$(ls "$GH_ARTIFACTS" | wc -l | sed 's/^ *//;s/ *$//')
           echo "count=$artifact_count" >> $GITHUB_OUTPUT
+        env:
+          GH_ARTIFACTS: ${{ runner.temp }}/gh-release-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
 
       - name: Publish artifacts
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -1533,6 +1534,32 @@ jobs:
           # Publish anything added to with the gh-release name
           files: ${{ runner.temp }}/gh-release-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}/*
 
+      - name: Download source artifact
+        if: steps.gh_artifacts.outputs.count != '0'
+        uses: actions/download-artifact@v3
+        with:
+          name: source-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
+          path: ${{ runner.temp }}
```
pipex, line 1542: Tried to use the syntax `<<: *downloadSourceArtifact` so I could just override with the `if`, but GH didn't seem to like it

flowzone.yml
```diff
@@ -1547,31 +1549,19 @@ jobs:
...
         if: needs.versioned_source.outputs.version != ''
-        shell: bash
         run: |
           set -ea
           # prevent git from existing with 141
           set +o pipefail
           previous_tag="$(git tag --sort=-version:refname | head -n 2 | tail -n 1)"
-          release_notes="$(git log ${previous_tag}..HEAD --pretty=reference)"
+          # Use HEAD~2 to skip the merge and versioning commit
+          release_notes="$(git log ${previous_tag}..HEAD~2 --pretty=reference)"
```
klutchell, line 1564: This isn't bad, but would it be even better to use the head commit id instead of `HEAD~2`?
```suggestion
          release_notes="$(git log ${previous_tag}..${{ github.event.pull_request.head.sha || github.event.head_commit.id }} --pretty=reference)"
```
Maybe you tried this already, but I think the above would always be the last commit from the PR?

Otherwise `HEAD~2` is fine and unlikely to cause problems.

pipex: That's smart, I didn't think of that. I'll make the change

pipex:
@balena-ci I self certify!

Issue #332, opened 2022-11-24
Avoid leaking secrets in arbitrary code


klutchell:
If we want to enable `pull_request_target` for external contributions with secrets, we need to avoid leaking those secrets.

In general, we need to sandbox all arbitrary code (npm, docker, etc), only expose secrets to trusted code (workflow files).
- [x] [skip npm login for installing private dependencies if PR is a fork](https://github.com/product-os/flowzone/pull/339)
- [x] [perform docker login steps AFTER running any docker builds and tests](https://github.com/product-os/flowzone/pull/335)
- [x] [exit with an error if PR is a fork and `COMPOSE_VARS` is set](https://github.com/product-os/flowzone/pull/336)
- [x] [exit with an error if PR is a fork and `.github/` files have been modified](https://github.com/product-os/flowzone/pull/336)
- [x] [skip custom actions if PR is a fork and `restrict_custom_actions` is `true`](https://github.com/product-os/flowzone/pull/337)

PR #334, opened 2022-11-24
Remove the global GHCR_TOKEN env var as it contains secrets

klutchell:
Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #335, opened 2022-11-24
Only login to docker registries immediately before pushing

klutchell:
This prevents docker credentials from leaking into arbitrary test code.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #336, opened 2022-11-24
Prevent usage of COMPOSE_VARS secret for external contributions

klutchell:
Since this secret is injected into the environment before running compose tests we cannot allow it to be used with external contributions.

Change-type: patch


.github/workflows/flowzone.yml
```diff
@@ -283,6 +283,24 @@ jobs:
         with:
           cmd: bash -c "echo $INPUT | tr -d '[:space:]'"
           separator: ","
+      - name: Check external contributor limitations
+        if: github.event.pull_request.head.repo.fork
```
ab77, line 287: @klutchell this will trip users who want to fork and run Flowzone independently from us (since it will always be true).

klutchell: Good catch

ab77 (approved):

PR #337, opened 2022-11-25
Skip custom actions for external contributions by default

klutchell:
This toggle can be changed via an input after the custom actions have been vetted for security and cannot leak secrets.

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -157,6 +157,11 @@ on:
         type: string
         required: false
         default: ""
+      restrict_custom_actions:
+        description: "Do not execute custom actions for external contributors. Only remove this restriction if custom actions have been vetted as secure."
+        type: boolean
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -582,7 +587,9 @@ jobs:
             echo "enabled=false" >> $GITHUB_OUTPUT
           fi
 
-      - name: Check for custom jobs
+      - name: Check for custom actions
+        # skip custom actions for external contributions by default
+        if: github.event.pull_request.head.repo.full_name == github.repository || inputs.restrict_custom_actions == false
```
klutchell, line 592: Maybe we should print a notice here instead of just quietly skipping the custom actions?

ab77 (approved):

PR #342, opened 2022-11-28
Send branch to cloudflare so merge to master deploys to prod

anujdeshpande:
Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1481,7 +1481,7 @@ jobs:
     name: Build website using Docusaurus
     runs-on: ${{fromJSON(inputs.runs_on)}}
     env:
-     CF_BRANCH: ${{ github.event.pull_request.head.ref }}
+     CF_BRANCH: ${{ github.event.pull_request.head.ref || github.event.repository.default_branch }}
     needs:
       - project_types
       - versioned_source
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1480,7 +1480,12 @@ jobs:
   website_build:
     name: Build website using Docusaurus
     runs-on: ${{fromJSON(inputs.runs_on)}}
-    needs: [project_types, versioned_source]
+    env:
+     CF_BRANCH: ${{ github.event.pull_request.head.ref }}
+    needs:
+      - project_types
+      - versioned_source
+      - event_types
     if: |
       needs.project_types.outputs.has_readme == 'true' &&
       inputs.cloudflare_website != ''
@@ -1497,12 +1502,16 @@ jobs:
           org: ${{ github.repository_owner }}
           default_branch: ${{ github.event.repository.default_branch }}
           url: https://${{ inputs.cloudflare_website }}.pages.dev
+      - name: Update deploy branch for merged PRs
+        if: needs.event_types.outputs.pr_merged == 'true'
+        run: |
+            echo "CF_BRANCH=${{ github.event.repository.default_branch }}" >> $GITHUB_ENV
       - name: Cloudflare Pages
         uses: cloudflare/wrangler-action@2.0.0
         with:
           apiToken: ${{ secrets.CF_API_TOKEN }}
           accountId: ${{ secrets.CF_ACCOUNT_ID }}
-          command: pages publish --project-name=${{ inputs.cloudflare_website }} build/ >> $GITHUB_STEP_SUMMARY
+          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} build/ >> $GITHUB_STEP_SUMMARY
   ###################################################
   # GitHub
   ###################################################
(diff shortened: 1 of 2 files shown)


anujdeshpande:
@balena-ci rebase

flowzone.yml
```diff
@@ -1486,12 +1489,26 @@ jobs:
...
+          echo $CF_BRANCH
+          echo "CF_BRANCH=${CF_BRANCH}" >> $GITHUB_ENV
+        shell: bash
       - name: Cloudflare Pages
         uses: cloudflare/wrangler-action@2.0.0
         with:
           apiToken: ${{ secrets.CF_API_TOKEN }}
           accountId: ${{ secrets.CF_ACCOUNT_ID }}
-          command: pages publish --project-name=${{ inputs.cloudflare_website }} build/ >> $GITHUB_STEP_SUMMARY
+          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} build/ >> $GITHUB_STEP_SUMMARY
```
ab77, line 1511: can you please break this onto multiple lines for readability using (e.g.):

```
command: |
  pages publish \
    --branch ${{ ...
```

ab77 (changes requested):
Kyle may have some real optimisation suggestions, otherwise LGTM.

flowzone.yml
```diff
@@ -1486,12 +1489,26 @@ jobs:
...
+          echo ${{ fromJSON(needs.event_types.outputs.pr_merged) }}
+          if [[ ${{ fromJSON(needs.event_types.outputs.pr_merged) }} == 'true' ]]
+          then
+            CF_BRANCH=${{ github.event.repository.default_branch }}
+          else
+            CF_BRANCH=${{ github.event.pull_request.head.ref }}
+          fi
+          echo $CF_BRANCH
+          echo "CF_BRANCH=${CF_BRANCH}" >> $GITHUB_ENV
+        shell: bash
```
klutchell, line 1505:
```suggestion
      - name: Update deploy branch for merged PRs
        if: needs.event_types.outputs.pr_merged == 'true'
        run: |
            echo "CF_BRANCH=${{ github.event.repository.default_branch }}" >> $GITHUB_ENV
```

Then you can just put a job-level default value for env:
```
  env:
     CF_BRANCH: github.event.pull_request.head.ref
```

flowzone.yml
```diff
@@ -1469,7 +1469,12 @@ jobs:
   website_build:
     name: Build website using Docusaurus
     runs-on: ${{fromJSON(inputs.runs_on)}}
-    needs: [project_types, versioned_source]
+    env:
+     CF_BRANCH: github.event.pull_request.head.ref
```
klutchell, line 1473: This job is not currently limited to pull requests, it will also run on push of tags (if enabled at the top level).

So we should either check that the event type is a pr, or decide what to do if someone has tagged a branch since `github.event.pull_request...` will not exist in that case.

flowzone.yml
anujdeshpande: Probably an issue with how the CF wrangler action handles the `command` input - but breaking it down into multiple lines doesn't work - 

see logs here - https://github.com/balena-io-hardware/meta-hardwareProcess-doc/actions/runs/3576740889/jobs/6015096551#step:9:85

code here - https://github.com/product-os/flowzone/pull/342/files#diff-d899bb0fa13938dbb09e245516c63598abe5837797ab4bdf9860e7f1b70f3cadR1503

Unless there's another way to do this, I guess we are stuck with one big line unfortunately.

flowzone.yml
anujdeshpande: I could add a check on this line @klutchell 
https://github.com/product-os/flowzone/pull/342/files#diff-d899bb0fa13938dbb09e245516c63598abe5837797ab4bdf9860e7f1b70f3cadR1479

Don't know much about tags - but I guess I could check that type = PullRequestEvent?
https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types

flowzone.yml
```diff
@@ -1469,7 +1469,12 @@ jobs:
   website_build:
     name: Build website using Docusaurus
     runs-on: ${{fromJSON(inputs.runs_on)}}
-    needs: [project_types, versioned_source]
+    env:
+     CF_BRANCH: ${{ github.event.pull_request.head.ref }}
```
klutchell, line 1473:
```suggestion
     CF_BRANCH: ${{ github.event.pull_request.head.ref || github.event.repository.default_branch }}
```
I think just having the default branch as a fallback here will handle tag events. We don't use this workflow internally but tags would be treated like a push to master.

anujdeshpande: @klutchell done! could you approve the PR please?

klutchell: @ab77 has changes requested, I can override his review but let's give him a chance to clear it himself. LGTM though!

vipulgupta2048:
Thanks for getting this through @anujdeshpande, This PR is blocking docs deployments for several repos. 
IMHO if @ab77's comment is not critical then let's dismiss the review and get this merged.

klutchell (approved):

PR #343, opened 2022-11-29
Standardise repository settings

ab77:


flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_enable_squash_merge:
+        description: "Enable merging pull requests via squashed commit"
+        type: boolean
+        required: false
+        default: false
+      repo_enable_wiki:
+        description: "Enable wiki in the repository"
+        type: boolean
+        required: false
+        default: false
```
Page-, line 244: Will disabling a wiki for an existing project lose info in that wiki?

flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_enable_projects:
+        description: "Enable projects in the repository"
+        type: boolean
+        required: false
+        default: false
+      repo_enable_rebase_merge:
+        description: "Enable merging pull requests via rebase"
+        type: boolean
+        required: false
+        default: true
```
Page-, line 234: I believe this needs to be false to work with flowzone

flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_enable_merge_commit:
+        description: "Enable merging pull requests via merge commit"
+        type: boolean
+        required: false
+        default: false
+      repo_enable_projects:
+        description: "Enable projects in the repository"
+        type: boolean
+        required: false
+        default: false
```
Page-, line 229: Will this lose info for repos currently using projects?

flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_enable_issues:
+        description: "Enable issues in the repository"
+        type: boolean
+        required: false
+        default: true
+      repo_enable_merge_commit:
+        description: "Enable merging pull requests via merge commit"
+        type: boolean
+        required: false
+        default: false
```
Page-, line 224: This is what we currently use everywhere and I think is required for flowzone

flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_description:
+        description: "Description of the repository"
+        type: string
+        required: false
+        default: ${{ github.repository }}
+      repo_homepage:
+        description: "Repository home page URL"
+        type: string
+        required: false
+        default: ""
```
Page-, line 209: Will this clobber the existing urls for repos that have them?

flowzone.yml
```diff
@@ -2084,3 +2154,46 @@ jobs:
           echo "result<<${DELIMITER}" >> $GITHUB_OUTPUT
           echo "${result}" >> $GITHUB_OUTPUT
           echo "${DELIMITER}" >> $GITHUB_OUTPUT
+
+  ###################################################
+  ## configure standard repository settings
+  ###################################################
```
klutchell, line 2160: Could this logic not be merged with the branch protection steps? Seems like they should fall into the same category and use some of the same code.

flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_delete_branch_on_merge:
+        description: "Delete head branch when pull requests are merged"
+        type: boolean
+        required: false
+        default: true
+      repo_description:
+        description: "Description of the repository"
+        type: string
+        required: false
+        default: ${{ github.repository }}
```
Page-, line 204: Will this lose existing descriptions?

klutchell:
This PR should be a minor change, if not our first major. Definitely bigger than a patch since it will immediately change the repository settings of anyone using it.

flowzone.yml
```diff
@@ -177,6 +177,76 @@ on:
...
+      repo_allow_forking:
+        description: "Allow forking of an organization repository"
+        type: boolean
+        required: false
+        default: true
+      repo_default_branch:
+        description: "Set the default branch name for the repository"
+        type: string
+        required: false
+        default: main
```
Page-, line 194: Does this play nicely with ones that currently use `master`? (I assume so because github has tried to make it easy, but worth ensuring)

Page-:
@klutchell even if we mark it as a major every repo will immediately pick it up anyway, but it's definitely at least a minor and not a patch as it does add new features you can use

flowzone.yml
ab77: I thought about that, but decided not to mix them. If you want, we can combine.

flowzone.yml
ab77: Same answer as for wiki.

flowzone.yml
ab77: It won't set an empty URL (default), but if the user sets a URL, it will set it and overwrite whatever is currently there.

flowzone.yml
ab77: We won't know until we start using it , I guess.

klutchell:
I noticed that the Flowzone repo settings reverted the **Always suggest updating pull request branches** to `false` when it was definitely true before. Is it possible that the repo edit command is reverting settings that aren't available on the CLI?

ab77:
> I noticed that the Flowzone repo settings reverted the **Always suggest updating pull request branches** to `false` when it was definitely true before. Is it possible that the repo edit command is reverting settings that aren't available on the CLI?

Maybe, or maybe I fat fingered it. In any case, we need to programatically set this as part of repository config using `gh api ...` request (need to figure out where this setting lives).

flowzone.yml
Page-: Can this logic instead use https://github.com/marketplace/actions/repo-settings-sync rather than our own thing?

ab77: Tough one.. Do we trust `kbrashears5`?

In any case, we would still have to pre-process the inputs, so it would be two steps. I am happy to refactor to use this external code and pin it to a commit if The People decide this is the way..

ab77:
> I noticed that the Flowzone repo settings reverted the **Always suggest updating pull request branches** to `false` when it was definitely true before. Is it possible that the repo edit command is reverting settings that aren't available on the CLI?

@klutchell probably that's what was happening. I found the key in the schema and it's now part of the config (default: true).

ab77:
@Page- @klutchell if there are no further objections, I'll merge this tomorrow.

flowzone.yml
klutchell: > Tough one.. Do we trust `kbrashears5`?
> 
> In any case, we would still have to pre-process the inputs, so it would be two steps. I am happy to refactor to use this external code and pin it to a commit if The People decide this is the way..

We could always pin to a commit hash to avoid the trust problem, but it looks like we are already supporting fields that the action does not, so I guess we will stick with our method.

flowzone.yml
```diff
@@ -2139,24 +2229,29 @@ jobs:
...
+          message="$(echo "${result}" | jq -r .message)"
+
+          if ! [[ $message =~ null ]]; then
+            echo ::error title={url}::"apply_branch_protection_rules failed with ${message} ${result} "
             exit 1
           fi
 
-          if [[  $(echo "$result" | jq '.required_pull_request_reviews.required_approving_review_count') -ne ${{ inputs.required_approving_review_count }} ]]; then
+          if [[ $(echo "$result" | jq '.required_pull_request_reviews.required_approving_review_count') -ne ${{ inputs.required_approving_review_count }} ]]; then
             echo ::error title={url}::"branch protection required_pull_request_reviews.required_approving_review_count not set to ${{ inputs.required_approving_review_count }}"
```
klutchell, line 2255: I know you didn't write this error, but is there something more helpful we can print here?

ab77: @klutchell I am struggling to understand what that entire block does:

```
...
          if [[ $(echo "$result" | jq '.required_pull_request_reviews.required_approving_review_count') -ne ${{ inputs.required_approving_review_count }} ]]; then
            echo ::error title={url}::"branch protection required_pull_request_reviews.required_approving_review_count not set to ${{ inputs.required_approving_review_count }}"
            exit 1
          fi

          DELIMITER=$(echo $RANDOM | md5sum | head -c 32)
          echo "result<<${DELIMITER}" >> $GITHUB_OUTPUT
          echo "${result}" >> $GITHUB_OUTPUT
          echo "${DELIMITER}" >> $GITHUB_OUTPUT
```

As fas as I can see, `DELIMITER` isn't used anywhere. Shall I just remove the whole lot?

klutchell (approved):

PR #344, opened 2022-11-29
Pin third-party actions to commits

maggie0002:
This PR demonstrates pinning third-party actions to commits. Pinning to a commit increases security by ensuring that only that code can ever be used inside Flowzone. Without specified commits, someone could compromise the account of the third-party, replace the release tag with malicious content and infiltrate Flowzone. 

More on GitHub Actions security can be found here: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

This PR only pins third-party workflows that were already pointing to a static semver (such as actions pointed at v1.2.3) and subsequently has no impact on operation. We should consider also pinning the other third-party actions, although it will mean that they do not receive automatic updates (such as the `@v3` tag which will get all v3 updates). 

Change-type: minor

### DIFF
--- a/.github/actions/publish/action.yml
+++ b/.github/actions/publish/action.yml
@@ -25,7 +25,7 @@ runs:
 
     # Test publishing release artifacts to github
     - name: Upload generated flowzone.yml
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb
       with:
         name: gh-release-${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
         path: .github/workflows/flowzone.yml
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -339,7 +339,7 @@ jobs:
       - name: Checkout merge branch
         if: github.event_name == 'pull_request_target' && needs.event_types.outputs.pr_closed != true
         id: checkout_merge
-        uses: actions/checkout@v3
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b
         with:
           fetch-depth: 0
           submodules: recursive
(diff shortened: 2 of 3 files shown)


ab77:
> As I mentioned above we have Renovate enabled on this org to bump actions dependencies.

I couldn't see it here: https://github.com/balena-io/renovate-config/blob/master/.github/renovate.json

klutchell:
> > As I mentioned above we have Renovate enabled on this org to bump actions dependencies.
> 
> I couldn't see it here: https://github.com/balena-io/renovate-config/blob/master/.github/renovate.json

I thought this org was included in https://github.com/product-os/renovate-config/blob/master/renovate.json

ab77:
Oh yes, of course - it's there. Completely different approach, using external renovate hosted bot. We should probably look into that soon.

klutchell:
soonTM

klutchell:
Looks like Renovate can update actions with a commit sha and they provide examples on how to enable it:
https://docs.renovatebot.com/modules/manager/github-actions/#additional-information

maggie0002:
> As for the other actions that were not even pinned to a semver, I think it's worth pinning those as well. At least to a full semver, if not a commit. 

If we are going to pin to a full semver we may as well pin to a commit, as the two would function the same way, and the commit is more secure. Assuming that doesn't lead to any problems with Renovate. 

That said, I wouldn't lose any sleep over trusted third-party actions, like the official Docker ones. It is just the ones that I hadn't heard of that stuck out for me (such as `dawidd6/action-download-artifact@v2`).

klutchell:
@maggie0002 pinning all actions to digests sounds good to me, we just need to add the following to our [renovate config](https://github.com/product-os/renovate-config/blob/master/renovate.json) first.
```
{
  "extends": ["helpers:pinGitHubActionDigests"]
}
```
Once that's done you can update this PR to pin all actions to digests and we can merge it.

maggie0002:
@klutchell, assuming all the tests pass, this should now be ok with all the actions pinned to commits. The only exceptions are our own maintained actions such as `product-os/docusaurus-theme@master`. I figure these should update automatically and as they are maintained by us and not third-parties, it makes sense for them to always be latest. It avoids us having to come over to this repo and merge the Renovate PR.

klutchell:
Looks like I found the first issue with Renovate and this repo. Renovate tries to update the workflow file directly and does not know how to change the template in the root of the project.

See these failed PRs:
- https://github.com/product-os/flowzone/pull/386
- https://github.com/product-os/flowzone/pull/387

klutchell:
Either way, it seems this PR is not necessary as Renovate will re-pin everything for us!

PR #346, opened 2022-11-30
For merged PR events we need the merged source

klutchell:
Otherwise we will fail to push the versioned commits to master.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #347, opened 2022-11-30
Use merge_commit_sha to checkout PR sources

klutchell:
Using the direct link to refs/pull does not work for merged PRs.

Change-type: patch
See: https://github.com/product-os/flowzone/pull/346


klutchell:
@balena-ci I self-certify!

PR #352, opened 2022-11-30
Fix refence to merge_commit_sha that was returning empty

klutchell:
Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #355, opened 2022-12-01
Allow GitHub pre-release(s)

ab77:
This change allows projects like Etcher to be configured to not make every release final, thus avoiding signalling potentially garbage/invalid releases to the world.

It is entirely possible for all Etcher tests to pass with flying colours (green that is) and still not work once installed on one or more platforms. One way to gain more confidence here is to expand Etcher test coverage, implementing proper e2e tests, but we are not there yet.

So, setting `github_prerelease: true` on Etcher repository will create GH releases and enable manual testing to take place before marking a release final/latest.

cc @aethernet @zwhitchcox @mcraa

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -157,6 +157,11 @@ on:
         type: string
         required: false
         default: ""
+      github_prerelease:
+        description: "Finalise releases on merge."
+        type: boolean
+        required: false
+        default: false
       restrict_custom_actions:
         description: "Do not execute custom actions for external contributors. Only remove this restriction if custom actions have been vetted as secure."
         type: boolean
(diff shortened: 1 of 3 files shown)


ab77:
Related to https://github.com/balena-io/etcher/issues/3853

flowzone.yml
```diff
@@ -157,6 +157,16 @@ on:
...
+        description: "Finalise releases on merge."
+        type: boolean
+        required: false
+        default: true
+      github_make_latest:
+        description: "Move latest pointer on merge."
+        type: boolean
+        required: false
+        default: true
       restrict_custom_actions:
```
pipex, line 170: These options seem overly specific to the github job implementation details. Maybe there is a way to merge them into one, e.g. `github_prerelease`, where `make_latest = !github_prerelease` or something like that?

klutchell (approved):

PR #356, opened 2022-12-02
Remove the checkout ref until it can be fixed properly

klutchell:
This will break external contributions but will
unblock internal tasks.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #359, opened 2022-12-02
Use verbose output in calls to `cross` command

pipex:
This will help with better identification of problems during rust builds

Change-type: patch


pipex:
@balena-ci I self certify!

PR #374, opened 2022-12-07
Enable PR auto-merge

ab77:
@klutchell this should be auto-merged as soon as it's approved (assuming it passes tests).

The only caveat I've found so far, is that if a PR is initially draft, but then marked ready, there needs to be a dummy commit pushed to force the workflow to run again. PRs can't be set to automatically merge unless they are ready for review.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -820,6 +820,20 @@ jobs:
             exit 1
           fi
 
+      # if auto_merge is enabled at the repository level, configure PR auto-merge
+      - name: Enable PR auto-merge
+        if: |
+          github.event.pull_request.draft == false &&
+          needs.event_types.outputs.do_draft == 'true' &&
+          inputs.repo_config == true
+        run: |
+          if [[ '${{ inputs.repo_enable_auto_merge }}' =~ true ]]; then
+              gh pr merge ${{ github.event.pull_request.number }} --merge --auto
+          fi
+
+        env:
+         GH_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
+
       - name: Import GPG key for signing commits
         if: inputs.disable_versioning != true && needs.event_types.outputs.pr == 'true'
         id: import-gpg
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -820,6 +820,23 @@ jobs:
...
+        if: |
+          github.event.pull_request.draft == false &&
+          needs.event_types.outputs.pr == 'true' &&
+          needs.event_types.outputs.do_draft == 'true' &&
+          inputs.repo_config == true
+        run: |
+          if [[ '${{ inputs.repo_enable_auto_merge }}' =~ true ]]; then
+              gh pr merge ${{ github.event.pull_request.number }} --merge --auto
+          else
+              gh pr merge ${{ github.event.pull_request.number }} --merge --disable-auto
```
klutchell, line 834: Do we need to disable auto merge if it's not enabled at the repo level (by Flowzone or otherwise)?

It seems that if the repo allows it, and a user enabled it for the PR, we might as well leave the setting alone.

flowzone.yml
```diff
@@ -820,6 +820,23 @@ jobs:
             exit 1
           fi
 
+      # if auto_merge is enabled at the repository level, configure PR auto-merge
+      - name: Enable PR auto-merge
+        if: |
+          github.event.pull_request.draft == false &&
+          needs.event_types.outputs.pr == 'true' &&
```
klutchell, line 827:
```suggestion
```
I don't think you need to check if it's a pr, since `do_draft` is only true for PRs. It doesn't hurt, but probably redundant.

klutchell (changes requested):

flowzone.yml
ab77: I was thinking this would allow toggling PR auto-merge as well as the global repo setting via the same `repo_enable_auto_merge` input. Maybe we need a separate input to control the PR setting @klutchell?

klutchell: @ab77 I don't think we need a new input for this. Enabling auto-merge on every PR as a Flowzone step makes sense, I just don't see any point in disabling auto-merge on any PR under any circumstances.

If I have a repo, and I've enabled auto merge at the repo level (not via Flowzone, manually since I'm admin), and I decide to toggle auto-merge for one of my PRs to ON, why would I ever want Flowzone to turn that off again?

My point is, regardless of the repo-level setting, auto merge is disabled on each PR by default, and has to be enabled intentionally, why would Flowzone ever turn it off?

If a repo admin didn't want PRs to use auto-merge it could be disabled at the repo level, with our without Flowzone.

ab77: > ... auto merge is disabled on each PR by default, and has to be enabled intentionally, why would Flowzone ever turn it off?

Yes, that makes sense. Since it's off by default, so there is no need to ever turn it off. I'll make the change.

klutchell (approved):

PR #377, opened 2022-12-08
PR auto-merge only if reviewers > 0

ab77:
Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -820,20 +820,6 @@ jobs:
             exit 1
           fi
 
-      # if auto_merge is enabled at the repository level, configure PR auto-merge
-      - name: Enable PR auto-merge
-        if: |
-          github.event.pull_request.draft == false &&
-          needs.event_types.outputs.do_draft == 'true' &&
-          inputs.repo_config == true
-        run: |
-          if [[ '${{ inputs.repo_enable_auto_merge }}' =~ true ]]; then
-              gh pr merge ${{ github.event.pull_request.number }} --merge --auto
-          fi
-
-        env:
-         GH_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
-
       - name: Import GPG key for signing commits
         if: inputs.disable_versioning != true && needs.event_types.outputs.pr == 'true'
         id: import-gpg
@@ -2147,6 +2133,9 @@ jobs:
       BRANCH_PROTECTION_URI: repos/${{ github.repository }}/branches/${{ github.event.repository.default_branch }}/protection
       GH_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
     steps:
+      - *downloadSourceArtifact
+      - *extractSourceArtifact
+
       # loop over all needed jobs and fail this job if any of them failed
       - name: Check needed jobs
         run: |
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -2278,7 +2280,7 @@ jobs:
     name: Standardise repository settngs
     runs-on: ${{ fromJSON(inputs.runs_on) }}
     timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
-    needs: [event_types]
+    needs: [protect_branch]
```
klutchell, line 2283: We can't change this to depend on `protect_branch` because protect branch only runs on unmerged PRs (so `do_final` would never be true)

I think we can leave this as-is?

flowzone.yml
```diff
@@ -2270,6 +2259,19 @@ jobs:
           echo "${result}" >> $GITHUB_OUTPUT
           echo "${DELIMITER}" >> $GITHUB_OUTPUT
 
+      # if auto_merge is enabled at the repository level, configure PR auto-merge
+      - name: Enable PR auto-merge
+        if: |
+          github.event.pull_request.draft == false &&
+          needs.event_types.outputs.do_draft == 'true' &&
```
klutchell, line 2266:
```suggestion
```
Don't need this check as protect branch already depends on do_draft.

flowzone.yml
```diff
@@ -2270,6 +2259,19 @@ jobs:
...
+          inputs.repo_config == true
```
klutchell, line 2267:
```suggestion
```
Might as well delete this check since we are effectively making this change if branch protection is true and we are on an active PR. Does not depend on repo config anymore.

flowzone.yml
```diff
@@ -2270,6 +2259,19 @@ jobs:
...
+        run: |
+          if [[ '${{ inputs.repo_enable_auto_merge }}' =~ true ]]; then
```
klutchell, line 2269:
```suggestion
```
This is a setting that may have been enabled at the repo level manually, outside the Flowzone context, so there's no need to check the input here. They might not be using flowzone for repo config.

We should either check the current setting on the repo itself, or just try to enable auto-merge and handle the failure gracefully.

klutchell (changes requested):

flowzone.yml
ab77: but if we remove this conditional, then auto-merge will be enabled by default with the next run..

flowzone.yml
ab77: I am using the same input to control both the repo level setting and per-PR setting. If we remove this conditional, auto-merge will be enabled always, unless required_approving_review_count is zero. Do we want this?

klutchell: I think if the repo allows it we should be able to toggle it on assuming required reviewers is more than one. So yes. It certainly makes more sense than re-using an optional input meant for applying repo settings on finalize.

Alternately, we could make a new input to control whether to try to enable the auto-merge toggle on the PR (and note that it will fail gracefully if not already allowed in the repo settings).

But if auto-merge is allowed on the repo, and the tests have passed, and at least one person has reviewed the PR, Bulldozer would have merged it anyway so why shouldn't Flowzone?

ab77: > But if auto-merge is allowed on the repo, and the tests have passed, and at least one person has reviewed the PR, Bulldozer would have merged it anyway so why shouldn't Flowzone?

That's a fair point.

flowzone.yml
```diff
@@ -2270,6 +2259,16 @@ jobs:
           echo "${result}" >> $GITHUB_OUTPUT
           echo "${DELIMITER}" >> $GITHUB_OUTPUT
 
+      # if auto_merge is enabled at the repository level, configure PR auto-merge
+      - name: Enable PR auto-merge
+        if: github.event.pull_request.draft == false
+        run: |
+          if [[ $(gh api '/repos/${{ github.repository }}' | jq -r .allow_auto_merge) =~ true ]]; then
```
klutchell, line 2266: This is nicer, I was just going to suggest `|| true` on the command below 😂

klutchell (approved):

PR #380, opened 2022-12-09
Add example config file

joshbwlng:
Change-type: patch
Signed-off-by: Josh Bowling <josh@monarci.com>

---

Add `example.yml` file for https://github.com/product-os/flowzonify to reference

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -1,0 +1,21 @@
+name: Flowzone
+
+on:
+  pull_request:
+    types: [opened, synchronize, closed]
+    branches: [main, master]
+  # allow external contributions to use secrets within trusted code
+  pull_request_target:
+    types: [opened, synchronize, closed]
+    branches: [main, master]
+
+jobs:
+  flowzone:
+    name: Flowzone
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
+    # prevent duplicate workflows and only allow one `pull_request` or `pull_request_target` for
+    # internal or external contributions respectively
+    if: |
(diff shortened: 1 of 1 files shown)


ab77:
@joshbwlng this would need to be synced up with the README.md (which already has this information). Do we add a lot of value by placing another file in there to keep in sync with the README manually? cc @klutchell

klutchell:
I agree with Anton that this is duplicating the information we would have to keep up-to-date. If we want an example file we should link to it from the README to avoid having the info in two places.

However I think what we have on the README is good enough?

ab77:
@joshbwlng I'll close this for now, if there is a strong reason for a separate file in the future, let's re-open then.

joshbwlng:
@klutchell @ab77 I had the same thought about it duplicating data before opening the PR (why I didn't merge it myself btw). The idea here is to enable flowzonify with https://github.com/product-os/flowzonify/pull/9 to be able to automatically open PRs on repos to update their flowzone config when necessary (think renovate/dependabot but for flowzone configs). For example to automatically open PRs to enable external contributions for our public repos, instead of having to do it manually. After thinking about it a bit more, I think it could also make sense to keep the reference in the flowzonify repo. This means we would need to manually update the reference in that repo, but it could then open all other PRs for us, much better than what we have now at least.

I had thought about moving the examples out of the flowzone readme into yaml files and then referencing them from the readme, but as an end-user I much prefer making examples like that immediately visible and available in the main readme. :thinking:

PR #395, opened 2022-12-14
Add input for custom website build

vipulgupta2048:
BalenaCloud docs is the special butterfly that would need this custom website build command to enable it's deployment on Cloudflare Pages using Flowzone. This is a patch until we can actually move the balenaCloud docs to Docusaurus but that will take a long time and this custom way of building websites could potenially be useful to others. 

Things to check 

- [x] @vipulgupta2048 will test the working balena-docs
- [x] Reviewers can check out the security implications of providing a free for all command to the users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
   - [Custom](#custom)
   - [Versioning](#versioning)
   - [Docs](#docs)
+  - [Website](#Website)
 - [Customization](#customization)
   - [Secrets](#secrets)
     - [`FLOWZONE_TOKEN`](#flowzone_token)
@@ -71,6 +72,8 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
     - [`custom_test_matrix`](#custom_test_matrix)
     - [`custom_publish_matrix`](#custom_publish_matrix)
     - [`custom_finalize_matrix`](#custom_finalize_matrix)
+    - [`cloudflare_website`](#cloudflare_website)
+    - [`docusaurus_website`](#docusaurus_website)
 
 - [Maintenance](#maintenance)
   - [Generate GPG keys](#generate-gpg-keys)
@@ -318,6 +321,13 @@ If you have [VersionBot3](https://github.com/apps/versionbot3) installed it will
 ### Docs
 
 If you have an `npm run doc` script then it will automatically be run and the generated `docs` folder will be published to a `docs` branch for github pages use.
+
+### Website
+
+If you have docs that you intend to publish on a website, checkout the [Getting Started](https://docusaurus-builder.pages.dev/) section of the Docusaurus builder. The docs will be built using the Docusaurus framework and published on Cloudflare Pages. 
+
+If you intend to use a custom framework for your docs build, then you can make use of the custom website build option by adding your desired build command in a input called `custom_website_build`. This command should generate your static site into a folder called `build` which will then be deployed to Cloudflare Pages. 
+
 
 ## Customization
 
@@ -669,6 +679,22 @@ Comma-delimited string of values that will be passed to the custom finalize acti
 Type: _string_
 
 Default: `''`
+
+#### `cloudflare_website`
+
+Setting this to your existing CF pages project name will generate and deploy a website. Skipped if empty.
+
+Type: _string_
+
+Default: `""`
+
+#### `docusaurus_website`
+
+Set to false to disable building a docusaurus website. If false the script `npm run deploy-docs` will be run if it exists.
+
+Type: _string_
+
+Default: true
 
 ## Maintenance
 
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -232,11 +232,11 @@ on:
         type: string
         required: false
         default: ""
-      custom_website_build:
-        description: "Add the custom command or script you want to run to build the website. Skipped if empty."
-        type: string
+      docusaurus_website:
+        description: "Set to false to disable building a docusaurus website. If false the script `npm run deploy-docs` will be run if it exists."
+        type: boolean
         required: false
-        default: ""
+        default: false
       github_prerelease:
         description: "Finalise releases on merge."
         type: boolean
@@ -708,12 +708,12 @@ jobs:
             echo "enabled=false" >> $GITHUB_OUTPUT
           fi
 
-      # Check if we have atleast a README to build a website from
+      # Check if we have at least a README to build a website from
 
       - name: Check for README for building a website
         id: has_readme
         run: |
-          if test -f "README.md"
+          if test -e "README.md"
           then
             echo "found README.md"
             echo "enabled=true" >> $GITHUB_OUTPUT
@@ -1607,10 +1607,15 @@ jobs:
     runs-on: ${{fromJSON(inputs.runs_on)}}
     env:
      CF_BRANCH: ${{ github.event.pull_request.head.ref || github.event.repository.default_branch }}
+
     needs:
       - project_types
       - versioned_source
       - event_types
+
+    if: |
+      inputs.cloudflare_website != ''
+
     steps:
       - *downloadSourceArtifact
       - *extractSourceArtifact
@@ -1622,8 +1627,8 @@ jobs:
       - name: Docusaurus Builder
         if: |
           needs.project_types.outputs.has_readme == 'true' &&
-          inputs.cloudflare_website != '' && inputs.custom_website_build = ''
-        uses: product-os/docusaurus-theme@master
+          inputs.docusaurus_website != false
+        uses: product-os/docusaurus-theme@9cb615b78bcf2c2f9d8fa683a296a52d7beefae1 # 1.3.0
         with:
           repo: ${{ github.event.repository.name }}
           org: ${{ github.repository_owner }}
@@ -1632,8 +1637,8 @@ jobs:
 
       - name: Custom Website Builder
         if: |
-          inputs.custom_website_build != ''
-        run: ${{ inputs.custom_website_build }}
+          inputs.docusaurus_website == false
+        run: npm run deploy-docs --if-present
 
       - name: Update deploy branch for merged PRs
         if: needs.event_types.outputs.pr_merged == 'true'
@@ -1645,7 +1650,7 @@ jobs:
         with:
           apiToken: ${{ secrets.CF_API_TOKEN }}
           accountId: ${{ secrets.CF_ACCOUNT_ID }}
-          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} $GITHUB_WORKSPACE/build/ >> $GITHUB_STEP_SUMMARY
+          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} build/ >> $GITHUB_STEP_SUMMARY
 
   ###################################################
   # GitHub
@@ -2267,7 +2272,7 @@ jobs:
             exit 1
           fi
 
-          # here the number of required approvals sent to the gh API is checked back from the PUT response from GH api 
+          # here the number of required approvals sent to the gh API is checked back from the PUT response from GH api
           # Specifies the number of reviewers required to approve pull requests. Use a number between 1 and 6 or 0 to not require reviewers (successful checks => merge) .
           if [[ $(echo "$result" | jq '.required_pull_request_reviews.required_approving_review_count') -ne ${{ inputs.required_approving_review_count }} ]]; then
             echo "::warning::Failed to set required approvers count. Check that branch protection is enabled on your repo and try again."
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -232,6 +232,11 @@ on:
         type: string
         required: false
         default: ""
+      custom_website_build:
+        description: "Add the custom command or script you want to run to build the website. Skipped if empty."
+        type: string
+        required: false
+        default: ""
       github_prerelease:
         description: "Finalise releases on merge."
         type: boolean
@@ -704,10 +709,11 @@ jobs:
           fi
 
       # Check if we have atleast a README to build a website from
+
       - name: Check for README for building a website
         id: has_readme
         run: |
-          if test -e "README.md"
+          if test -f "README.md"
           then
             echo "found README.md"
             echo "enabled=true" >> $GITHUB_OUTPUT
@@ -1605,9 +1611,6 @@ jobs:
       - project_types
       - versioned_source
       - event_types
-    if: |
-      needs.project_types.outputs.has_readme == 'true' &&
-      inputs.cloudflare_website != ''
     steps:
       - *downloadSourceArtifact
       - *extractSourceArtifact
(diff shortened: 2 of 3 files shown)


flowzone.yml
```diff
@@ -232,6 +232,11 @@ on:
         type: string
         required: false
         default: ""
+      custom_website_build:
+        description: "Add the custom command or script you want to run to build the website. Skipped if empty."
+        type: string
+        required: false
+        default: ""
```
klutchell, line 239: This input also needs to be documented on the README.

flowzone.yml
```diff
@@ -1617,12 +1620,20 @@ jobs:
...
           repo: ${{ github.event.repository.name }}
           org: ${{ github.repository_owner }}
           default_branch: ${{ github.event.repository.default_branch }}
-          url: https://${{ inputs.cloudflare_website }}.pages.dev
+          url: https://${{ inputs.cloudflare_website }}.pages.dev/
+
+      - name: Custom Website Builder
+        if: |
+          inputs.custom_website_build != ''
+        run: ${{ inputs.custom_website_build }}
```
klutchell, line 1636:
```suggestion
        run: npm run doc --if-present
```
Why not reuse the same npm script we support here?
https://github.com/product-os/flowzone/blob/740bed380dbaf3c3619bb470e75901a4ae8eb5bc/flowzone.yml#L1029

Then we could avoid a new input and just call an npm script if it exists.

It doesn't even need to be `run doc` but something appropriate we could call if it exists rather than a free form command that is prone to escape and special character issues.

flowzone.yml
```diff
@@ -1634,7 +1645,7 @@ jobs:
         with:
           apiToken: ${{ secrets.CF_API_TOKEN }}
           accountId: ${{ secrets.CF_ACCOUNT_ID }}
-          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} build/ >> $GITHUB_STEP_SUMMARY
+          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} $GITHUB_WORKSPACE/build/ >> $GITHUB_STEP_SUMMARY
```
klutchell, line 1648:
```suggestion
          command: pages publish --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} ${{ inputs.working_directory }}/build/ >> $GITHUB_STEP_SUMMARY
```

klutchell (changes requested):
I don't see this introducing any major security issues. Running arbitrary code as a build step is not ideal but we have to do it in several places. We just avoid having secrets exposed to the environment at that time.

vipulgupta2048:
Site works fine o fine: https://19331bda.balenacloud-docs.pages.dev/reference/hardware/wifi-dongles/
PR: https://github.com/balena-io/docs/pull/2459

flowzone.yml
vipulgupta2048: I saw that, and `npm run doc` deploys to GitHub pages. Hence, chose to make my own that deploys to CF pages and keep the 2 separate. I will look into the implementation if that's not true.

flowzone.yml
vipulgupta2048: Yeah that would be better if working_directory input defaults to something sensible.

flowzone.yml
klutchell: My point is an optional npm script is preferable over a new free-form command input. You can call the script whatever you would like and Flowzone can be configured to run it if it exists.

vipulgupta2048: Right on, but aren't we then blocking anyone else who is not using an NPM repo but wants to build their own site? 
The free-form command although rogue serves more people's preferences.

klutchell: Do we have any examples of building sites that do not use npm?

I think a free-form field is more likely to introduce issues with quoting, escaping characters, etc.

vipulgupta2048: > Do we have any examples of building sites that do not use npm?

Not at the moment. 

> I think a free-form field is more likely to introduce issues with quoting, escaping characters, etc.

Wouldn't those issues be issues of the users who are implementing it on their repository? Flowzone is providing a custom command. It will run what you provide essentially.

klutchell: We want this project to be opinionated, I think allowing any arbitrary shell command is asking for trouble, when we can assume 99% of use cases would work with an npm script.

vipulgupta2048: I agree on that one, I will get this setup as `npm run deploy` or actually go look into `npm run doc` first

flowzone.yml
```diff
@@ -704,10 +709,11 @@ jobs:
           fi
 
       # Check if we have atleast a README to build a website from
```
ab77, line 711: space missing

flowzone.yml
```diff
@@ -704,10 +709,11 @@ jobs:
...
+
       - name: Check for README for building a website
         id: has_readme
         run: |
-          if test -e "README.md"
+          if test -f "README.md"
```
ab77, line 716: any reason to require a file? might be a symlink?

flowzone.yml
```diff
@@ -1617,12 +1620,20 @@ jobs:
...
+        if: |
+          needs.project_types.outputs.has_readme == 'true' &&
+          inputs.cloudflare_website != '' && inputs.custom_website_build == ''
+        uses: product-os/docusaurus-theme@master
         with:
           repo: ${{ github.event.repository.name }}
           org: ${{ github.repository_owner }}
           default_branch: ${{ github.event.repository.default_branch }}
-          url: https://${{ inputs.cloudflare_website }}.pages.dev
+          url: https://${{ inputs.cloudflare_website }}.pages.dev/
```
ab77, line 1631: @vipulgupta2048 we are still limited to `.pages.dev` here. What if a user wants `https://flowzone.foo.bar`?

flowzone.yml
ab77: Do we have a use case right now for a "custom" web site build? If not, perhaps park it until we do and focus on a custom domain (instead of page.dev), which is likely to be the more frequent use case. But maybe I a missing something..

ab77 (changes requested):

flowzone.yml
vipulgupta2048: Considering what we talked about above @klutchell ^ Should we use this new input `inputs.custom_website_build`  being defined as a boolean instead for Flowzone to know what to do? 

If input present then 
    Custom
if not then 
    default?

vipulgupta2048: @ab77 My PR description describes why we need a custom website build. It's to unblock balenaCloud docs deployment to CF pages.

flowzone.yml
vipulgupta2048: @anujdeshpande is will be opening a PR about it very soon. Already planned and discussed.

flowzone.yml
vipulgupta2048: This change might have snuck in. Reverting.

flowzone.yml
vipulgupta2048: Do you mean the linebreak? after it? I didn't get this.

flowzone.yml
ab77: Thanks @vipulgupta2048. I also think using npm workflow instead of arbitrary commands is the more correct way to do this and perhaps this workflow can be combined with the existing `npm docs ..`? Though I am not sure what use case that handles. Maybe a different one.

vipulgupta2048: After reviewing that, I have started a conversation about this here: https://balena.zulipchat.com/#narrow/stream/348930-product-os.2Fflowzone/topic/Multiple.20project_types

I think we should drop the existing npm run docs functionality and resolve to using CF pages for docs. I can do that in a different PR. For now , 

I am choosing to go with a custom input that decides whether or not we are going custom and if we are then we run `npm run deploy`

vipulgupta2048:
Ready to be re-reviewed

flowzone.yml
vipulgupta2048: Yeah, this won't work. It fails with: 

`✘ [ERROR] ENOENT: no such file or directory, scandir '/github/workspace/tests/build'`

This could be because we change directories to build the docs in a clean directory. Reverting this back to the original.

flowzone.yml
```diff
@@ -232,6 +232,11 @@ on:
         type: string
         required: false
         default: ""
+      custom_website_build:
+        description: "Runs a custom command to build the website. The command needs to be in a npm script stored in `npm run deploy`. Skipped if empty or not true."
+        type: boolean
+        required: false
+        default: false
```
klutchell, line 239: I would much rather a toggle like `docusaurus_website` boolean and we could default to `true`.

flowzone.yml
```diff
@@ -1617,12 +1620,20 @@ jobs:
           node-version: 18
 
       - name: Docusaurus Builder
-        uses: product-os/docusaurus-builder@master
+        if: |
+          needs.project_types.outputs.has_readme == 'true' &&
+          inputs.cloudflare_website != '' && inputs.custom_website_build != 'true'
+        uses: product-os/docusaurus-theme@master
```
klutchell, line 1626: We should pin this to a version. Renovate will update it as needed.

flowzone.yml
```diff
@@ -1605,9 +1611,6 @@ jobs:
       - project_types
       - versioned_source
       - event_types
-    if: |
-      needs.project_types.outputs.has_readme == 'true' &&
-      inputs.cloudflare_website != ''
```
klutchell, line 1610: We still need to check that `inputs.cloudflare_website` is not empty here, because both docusaurus and custom still push to CF.

flowzone.yml
```diff
@@ -1617,12 +1620,20 @@ jobs:
           node-version: 18
 
       - name: Docusaurus Builder
-        uses: product-os/docusaurus-builder@master
+        if: |
+          needs.project_types.outputs.has_readme == 'true' &&
+          inputs.cloudflare_website != '' && inputs.custom_website_build != 'true'
```
klutchell, line 1625:
```suggestion
          inputs.docusaurus_website == true
```
We should check `inputs.cloudflare_website` at the top of the job as per my above comment, and use a docusaurus boolean here.

flowzone.yml
```diff
@@ -1617,12 +1620,20 @@ jobs:
...
         with:
           repo: ${{ github.event.repository.name }}
           org: ${{ github.repository_owner }}
           default_branch: ${{ github.event.repository.default_branch }}
-          url: https://${{ inputs.cloudflare_website }}.pages.dev
+          url: https://${{ inputs.cloudflare_website }}.pages.dev/
+
+      - name: Custom Website Builder
+        if: |
+          inputs.custom_website_build == 'true'
```
klutchell, line 1635:
```suggestion
          inputs.docusaurus_website == false
```

klutchell (changes requested):

flowzone.yml
ab77: > I am choosing to go with a custom input that decides whether or not we are going custom and if we are then we run npm run deploy

@vipulgupta2048 `deploy` script as in `npm run deploy` seems ambigious. Deploy what? `docs` seems more clear..

cc @klutchell

flowzone.yml
vipulgupta2048: Yeah, sounds good.

flowzone.yml
vipulgupta2048: Okay yeah that's a good catch

flowzone.yml
vipulgupta2048: After reading your other comments, I get what you mean. Yeah. Let's do that. Cool

flowzone.yml
vipulgupta2048: Do you mean `npm run docs` or `npm run deploy-docs` @ab77 because then the former runs the risk of sounding ambigous? 

Just to say again, this is only for one repository atm. Anyone else using it would be suspicious and needs to prove that they are indeed special or not. Mentioning this to prevent bikeshedding.

vipulgupta2048:
I think this is ready to go: @klutchell @ab77 How about some re-reviews.

README.md
```diff
@@ -71,6 +72,8 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
     - [`custom_test_matrix`](#custom_test_matrix)
     - [`custom_publish_matrix`](#custom_publish_matrix)
     - [`custom_finalize_matrix`](#custom_finalize_matrix)
+    - [`cloudflare_website`](#cloudflare_website)
+    - [`custom_website_build`](#custom_website_build)
```
klutchell, line 76:
```suggestion
    - [`docusaurus_website `](#docusaurus_website)
```

flowzone.yml
```diff
@@ -232,6 +232,11 @@ on:
         type: string
         required: false
         default: ""
+      docusaurus_website:
+        description: "Runs a custom command to build the website. The command needs to be in a npm script stored in `npm run deploy-docs`. Skipped if empty or not true."
```
klutchell, line 236: This description needs to reflect that it enables docusaurus, and if disabled it will run an npm script instead.

flowzone.yml
```diff
@@ -1617,12 +1625,20 @@ jobs:
...
         with:
           repo: ${{ github.event.repository.name }}
           org: ${{ github.repository_owner }}
           default_branch: ${{ github.event.repository.default_branch }}
-          url: https://${{ inputs.cloudflare_website }}.pages.dev
+          url: https://${{ inputs.cloudflare_website }}.pages.dev/
+
+      - name: Custom Website Builder
+        if: |
+          inputs.docusaurus_website == 'true'
```
klutchell, line 1640:
```suggestion
          inputs.docusaurus_website != true
```
I think inputs are actual booleans, and not strings like outputs.

flowzone.yml
```diff
@@ -1617,12 +1625,20 @@ jobs:
           node-version: 18
 
       - name: Docusaurus Builder
-        uses: product-os/docusaurus-builder@master
+        if: |
+          needs.project_types.outputs.has_readme == 'true' &&
+          inputs.docusaurus_website != 'true'
```
klutchell, line 1630:
```suggestion
          inputs.docusaurus_website == true
```

klutchell (changes requested):

README.md
```diff
@@ -670,6 +680,22 @@ Type: _string_
...
+
+Setting this to your existing CF pages project name will generate and deploy a website. Skipped if empty.
+
+Type: _string_
+
+Default: `""`
+
+#### `docusaurus_website`
+
+Set to false to disable building docusaurus website. Set the build command in a npm script stored in `npm run deploy-docs`. Runs docusaurus builder if empty.
```
klutchell, line 693:
```suggestion
Set to false to disable building a docusaurus website. If false the script `npm run deploy-docs` will be run if it exists.
```

klutchell (changes requested):

flowzone.yml
vipulgupta2048: Really, I see strings around all booleans everywhere for some reason.

klutchell (approved):

PR #399, opened 2022-12-16
Adds an option to set PRs to auto-merge.

cmfcruz:
When enabled, this job does the following:
- Sets the PR to automatically merge when all tests pass.
- Sets the default merge option for the PR.

Change-type: minor
Signed-off-by: Carlo Miguel F. Cruz <carloc@balena.io>

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -108,6 +108,16 @@ on:
         type: boolean
         required: false
         default: true
+      pr_enable_auto_merge:
+        description: "Set to true to enforce automatic merging after a successful run"
+        type: boolean
+        required: false
+        default: false
+      pr_merge_option:
+        description: "Set to enforce the default merge option for pull requests"
+        type: boolean
+        required: false
+        default: merge
       repo_config:
         description: Set to true to standardise repository settings after a successful run
         type: boolean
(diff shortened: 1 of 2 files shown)


cmfcruz:
This job currrently fails with the following error message:
```
no pull requests found for branch <PR URL>
```

Sample run here: https://github.com/balena-io/environment-staging/actions/runs/3711626801/jobs/6292255611

The problem seems to be related to [this open Github CLI issue](https://github.com/cli/cli/issues/575).

cmfcruz:
I worked around the GH CLI issue by downloading the versioned source and using that with the GH CLI to know which PR to work with.

cmfcruz:
@ab77 @klutchell @Page- , do you guys think we can include this job as part of flowzone or should I just create a separate GH action just to set auto-merge for every new PR?

klutchell (changes requested):
@cmfcruz we already do some of this during the protect branch step
https://github.com/product-os/flowzone/blob/dbc94fff9234149057a651a1c1e2e111875c16a5/flowzone.yml#L2270-L2278

cmfcruz:
I didn't see that bit.  Thanks for letting me know, @klutchell ! I'll close this PR and took note of that part in case we need to adjust something.

PR #401, opened 2022-12-17
Enable docusaurus website build by default

klutchell:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -217,7 +217,7 @@ on:
         description: Set to false to disable building a docusaurus website. If false the script `npm run deploy-docs` will be run if it exists.
         type: boolean
         required: false
-        default: false
+        default: true
       github_prerelease:
         description: Finalise releases on merge.
         type: boolean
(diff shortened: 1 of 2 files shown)


klutchell:
@balena-ci I self-certify!

vipulgupta2048:
I don't know if we do want any builds running until they have `cloudflare_website` input set

klutchell:
> I don't know if we do want any builds running until they have `cloudflare_website` input set

`cloudflare_website` is still required as per
https://github.com/product-os/flowzone/blob/5db785a6485d6f163f813dc5767179eb9b6eced6/flowzone.yml#L1617

But I missed this on your original PR, we want to default to docusaurus, not custom build script. Your PR actually failed the e2e test because this repo is configured for docusaurus but we accidentally defaulted to false.

vipulgupta2048:
The `docusaurus_website` being true and false is an implementation detail only valid for one repository atm. Can you refer to what does `default: true` does in GitHub actions

klutchell:
If you don't provide that input when calling Flowzone, it will default to
true. This is what we want because before your PR that input didn't exist
and docusaurus was assumed true in all cases.

Now we have that input so repos with custom docs builds can manually set
docusaurus_website to false.

vipulgupta2048:
Yeah, I edited the message a bit once I looked at the commit more closely. I will open a followup PR to fix the README

PR #405, opened 2022-12-26
Deploy to domain instead of pages.dev URL

anujdeshpande:
Change-type: patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,8 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
     - [`custom_test_matrix`](#custom_test_matrix)
     - [`custom_publish_matrix`](#custom_publish_matrix)
     - [`custom_finalize_matrix`](#custom_finalize_matrix)
-    - [`cloudflare_website`](#cloudflare_website)
+    - [`cloudflare_website`](#cloudflare_website
+    - [`domain_website`](#domain_website)
     - [`docusaurus_website`](#docusaurus_website)
 
 - [Maintenance](#maintenance)
@@ -683,6 +684,14 @@ Default: `''`
 #### `cloudflare_website`
 
 Setting this to your existing CF pages project name will generate and deploy a website. Skipped if empty.
+
+Type: _string_
+
+Default: `""`
+
+#### `domain_website`
+
+Set this to the domain that you want to deploy your website to. By default it will be deployed to a *.pages.dev
 
 Type: _string_
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
     - [`custom_test_matrix`](#custom_test_matrix)
     - [`custom_publish_matrix`](#custom_publish_matrix)
     - [`custom_finalize_matrix`](#custom_finalize_matrix)
-    - [`cloudflare_website`](#cloudflare_website
+    - [`cloudflare_website`](#cloudflare_website)
     - [`domain_website`](#domain_website)
     - [`docusaurus_website`](#docusaurus_website)
 
--- a/README.md
+++ b/README.md
@@ -691,7 +691,7 @@ Default: `""`
 
 #### `domain_website`
 
-Set this to the domain that you want to deploy your website to. By default it will be deployed to a *.pages.dev
+Set this to the domain that you want to deploy your website to. By default, it will be deployed to a `PROJECT_NAME.pages.dev` Cloudflare Pages URL.
 
 Type: _string_
 
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -264,6 +264,11 @@ on:
         type: string
         required: false
         default: ""
+      domain_website:
+        description: "Set this to the domain that you want to deploy your website to. By default it will be deployed to a *.pages.dev"
+        type: string
+        required: false
+        default: ""
       docusaurus_website:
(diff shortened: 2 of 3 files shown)


anujdeshpande:
The configuration of the domain is independent of the ability to push to it. So I could push to "my-project" expecting it to publish to my-project.pages.dev but if there's a domain configured for that project, say myproject.com, that would also get deployed.

We'd previously discussed about using a separate api key for repos that push to a domain. One of the reasons was so that no one accidentally pushes to a domain by simply copying a flowzone file (or does this maliciously)

I don't see how that will help solve the issue

Right now we have keys that are configured in our GitHub orgs that allow us any repo in that org to publish to any cloudflare pages project. So we would need to create repo specific keys that have access to only one pages project.

This makes org wide configuration of CF keys pointless

What else could we do? Am I missing something here?

README.md
```diff
@@ -72,7 +72,8 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
     - [`custom_test_matrix`](#custom_test_matrix)
     - [`custom_publish_matrix`](#custom_publish_matrix)
     - [`custom_finalize_matrix`](#custom_finalize_matrix)
-    - [`cloudflare_website`](#cloudflare_website)
+    - [`cloudflare_website`](#cloudflare_website
```
vipulgupta2048, line 75:
```suggestion
    - [`cloudflare_website`](#cloudflare_website)
```

README.md
```diff
@@ -688,6 +689,14 @@ Type: _string_
 
 Default: `""`
 
+#### `domain_website`
+
+Set this to the domain that you want to deploy your website to. By default it will be deployed to a *.pages.dev
```
vipulgupta2048, line 694:
```suggestion
Set this to the domain that you want to deploy your website to. By default, it will be deployed to a `PROJECT_NAME.pages.dev` Cloudflare Pages URL.
```

flowzone.yml
```diff
@@ -264,6 +264,11 @@ on:
         type: string
         required: false
         default: ""
+      domain_website:
+        description: "Set this to the domain that you want to deploy your website to. By default it will be deployed to a *.pages.dev"
```
vipulgupta2048, line 268: If you accept the description above then this could change as well

README.md
```diff
@@ -688,6 +689,14 @@ Type: _string_
...
+
+Type: _string_
+
+Default: `""`
```
vipulgupta2048, line 698: Shouldn't the default value go here

anujdeshpande:
@vipulgupta2048 see the comment above

README.md
vipulgupta2048: From your message above @anujdeshpande could someone provide a more generic input like `balena.io/docs` or `docs.balena.io` to configure both domain and BaseURL together for docusaurus. 

I know configuring docusaurus is easy but the hard part would be writing the scripts that would end up changing CNAME fields of URL entry in Cloudflare DNS. We could do the hard part manually or later but still give folks the option at least to define both URL (balena.io) + baseURL (/docs)

Example: Input can be `balena.io/cds/docs` = balena.io + /cds/docs 
This will make this input more versatile. I am asking this because I would need this for balenaCloud docs deployment: https://github.com/balena-io/docs/actions/runs/3694510530

This PR will fail because the base URL will be incorrect in docusaurus config.

vipulgupta2048: 1. We can easily use a node URL library on the docusaurusify script side to parse the URL input. 
2. Figure out domain and subdomain 
3. End up configuring the correct values. 
4. Then manually configuring Cloudflare DNS ourselves.

README.md
```diff
@@ -688,6 +689,14 @@ Type: _string_
 
 Default: `""`
 
+#### `domain_website`
+
+Set this to the domain that you want to deploy your website to. By default, it will be deployed to a `PROJECT_NAME.pages.dev` Cloudflare Pages URL.
```
ab77, line 694: @anujdeshpande Having two inputs makes little sense. Can we not use the same parameter for both? (i.e if the parameter doesn't include `.` dots, then assume it's `${{ website }}.pages.dev`, otherwise assume a FQDN.

ab77 (changes requested):

README.md
anujdeshpande: @ab77 That would add complexity to the action as we would have to figure out the pages project name from the URL. We use the official Cloudflare action which uses their wrangler CLI - and that only allows one to do `wrangler pages publish --project-name $PROJECT_NAME`

https://developers.cloudflare.com/workers/wrangler/commands/#publish-1

So basically flowzone/docusaurus action would have to 
- figure out it's a FQDN
- (using cURL)list all the pages project and their domains 
- find the project name from that list

Or we have a lookup table somewhere (but that's more complicated than this)

vipulgupta2048:
> @vipulgupta2048 see the comment above

@anujdeshpande Can you provide a link to which comment you are exactly referring to?

README.md
vipulgupta2048: > @vipulgupta2048 see the comment above

@anujdeshpande Can you provide a link to which comment you are exactly referring to?

anujdeshpande: @vipulgupta2048 https://github.com/product-os/flowzone/pull/405#issuecomment-1365349994

README.md
ab77: @anujdeshpande you can just run a bash step before where you do all the figuring out and provide and formatted output for the next step to consume.

PR #411, opened 2023-01-03
Check package architecture(s)

ab77:
Don't `run npm i|ci` when the platform isn't Linux.
(e.g.) https://github.com/balena-io-modules/winusb-driver-generator/blob/master/package.json#L30-L32

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -488,13 +488,22 @@ jobs:
       - name: Check for package.json
         id: npm
         run: |
+          runner_os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
           if test -f "package.json"
           then
             echo "found package.json"
-            echo "enabled=true" >> $GITHUB_OUTPUT
-            echo "private=$(jq -r '.private' package.json)" >> $GITHUB_OUTPUT
-            echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> $GITHUB_OUTPUT
-            echo "NODE_VERSIONS=[]" >> $GITHUB_ENV
+            os_count="$(jq '.os | length' package.json)"
+            index="$(jq --arg os "${runner_os}" '.os | index($os) | select( . != null )' package.json)"
+            if [[ -n "$index" ]] || [[ "$os_count" -lt 1 ]]
+            then
+                echo "found supported OS"
+                echo "enabled=true" >> $GITHUB_OUTPUT
+                echo "private=$(jq -r '.private' package.json)" >> $GITHUB_OUTPUT
+                echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> $GITHUB_OUTPUT
+                echo "NODE_VERSIONS=[]" >> $GITHUB_ENV
+            else
+                echo "$runner_os is not supported in package.json"
+            fi
           else
             echo "enabled=false" >> $GITHUB_OUTPUT
           fi
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -417,13 +417,22 @@ jobs:
       - name: Check for package.json
         id: npm
         run: |
+          runner_os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
           if test -f "package.json"
           then
             echo "found package.json"
-            echo "enabled=true" >> $GITHUB_OUTPUT
-            echo "private=$(jq -r '.private' package.json)" >> $GITHUB_OUTPUT
-            echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> $GITHUB_OUTPUT
-            echo "NODE_VERSIONS=[]" >> $GITHUB_ENV
+            os_count="$(jq '.os | length' package.json)"
+            index="$(jq --arg os "${runner_os}" '.os | index($os) | select( . != null )' package.json)"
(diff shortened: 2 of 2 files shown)


flowzone.yml
```diff
@@ -999,10 +999,14 @@ jobs:
 
       - name: Install dependencies
         run: |
-          if [ -e package-lock.json ] || [ -e npm-shrinkwrap.json ]; then
-            npm ci
-          else
-            npm i
+          runner_os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
+          os="$(cat < package.json | yq e .os[])"
```
klutchell, line 1003:
```suggestion
          exists="$(jq --arg os "$runner_os" '.os | index($os)' package.json)"
```

ab77: it only makes sense to compare runner_os with os key for value `linux`. That's the only convergence, between the two sets of values. There are more verbose way of checking, or we can just limit the npm steps to run `if: runner_os == 'Linux'`.

flowzone.yml
```diff
@@ -488,13 +488,19 @@ jobs:
       - name: Check for package.json
         id: npm
         run: |
+          runner_os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
           if test -f "package.json"
```
ab77, line 492: @klutchell moved higher up..

flowzone.yml
```diff
@@ -488,13 +488,20 @@ jobs:
...
             echo "found package.json"
-            echo "enabled=true" >> $GITHUB_OUTPUT
-            echo "private=$(jq -r '.private' package.json)" >> $GITHUB_OUTPUT
-            echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> $GITHUB_OUTPUT
-            echo "NODE_VERSIONS=[]" >> $GITHUB_ENV
+            os="$(cat < package.json | yq e .os[])"
+            index="$(jq --arg os "${runner_os}" '.os | index($os) | select( . != null )' package.json)"
+            if [[ -n "$index" ]] || [[ -z "$os" ]]
+            then
+                echo "found supported OS"
```
klutchell, line 499:
```suggestion
                echo "$runner_os is supported in package.json"
```

flowzone.yml
```diff
@@ -488,13 +488,20 @@ jobs:
...
+                echo "enabled=true" >> $GITHUB_OUTPUT
+                echo "private=$(jq -r '.private' package.json)" >> $GITHUB_OUTPUT
+                echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> $GITHUB_OUTPUT
+                echo "NODE_VERSIONS=[]" >> $GITHUB_ENV
+            fi
```
klutchell, line 504:
```suggestion
            else
                echo "$runner_os is not supported in package.json"
            fi
```

flowzone.yml
```diff
@@ -488,13 +488,20 @@ jobs:
...
         run: |
+          runner_os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
           if test -f "package.json"
           then
             echo "found package.json"
-            echo "enabled=true" >> $GITHUB_OUTPUT
-            echo "private=$(jq -r '.private' package.json)" >> $GITHUB_OUTPUT
-            echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> $GITHUB_OUTPUT
-            echo "NODE_VERSIONS=[]" >> $GITHUB_ENV
+            os="$(cat < package.json | yq e .os[])"
```
klutchell, line 495: > it only makes sense to compare runner_os with os key for value linux. That's the only convergence, between the two sets of values. There are more verbose way of checking, or we can just limit the npm steps to run if: runner_os == 'Linux'.

Not sure if you accidentally deleted this comment, but I'm not sure I follow.

In package.json `os` is an array of strings. It could be `["windows"]` or `["linux"]` or `["windows", "linux", "macos"]`. If ANY of them match the current runner OS (only linux for now, but eventually more) we would want to run the npm scripts.

klutchell (dismissed):

flowzone.yml
ab77: @klutchell I think we are good here, right? Current logic runs npm workflow if either the `os` key is omitted in `package.json` (i.e. most cases), or exists and matches `runner_os`.

klutchell: Oh I see, you were checking for an empty array, which I forgot to do.

flowzone.yml
```diff
@@ -488,13 +488,22 @@ jobs:
...
```
klutchell, line 495:
```suggestion
            os_count="$(jq '.os | length' package.json)"
```
Nitpick, but why use yq to check the value of a json structure?

flowzone.yml
```diff
@@ -488,13 +488,22 @@ jobs:
...
+            index="$(jq --arg os "${runner_os}" '.os | index($os) | select( . != null )' package.json)"
+            if [[ -n "$index" ]] || [[ -z "$os" ]]
```
klutchell, line 497:
```suggestion
            if [[ -n "$index" ]] || [[ "$os_count" -lt 1 ]]
```

flowzone.yml
ab77: Yeah, that's because I am an idiot and constantly mix YAML and JSON. Thank you.

klutchell (commented):
squash the commits and then LGTM

ab77:
@balena-ci I self-certify!

PR #412, opened 2023-01-04
Revert previous

ab77:
Otherwise the entire NPM pipeline is disabled and any assets built in custom workflows can't be published.


ab77:
@balena-ci I self-certify!

PR #423, opened 2023-01-13
Improve regex for matching required status checks

klutchell:
Allow policy-bot to be matched without the hyphen, and use case-insensitive matching for flowzone, policy-bot, and resinci when modifying the list of required status checks.

Change-type: patch


klutchell:
I self-certify!

PR #424, opened 2023-01-14
ci: Add needs to custom composite actions

issenn:
Change-type: minor

### DIFF
--- a/.github/actions/always/action.yml
+++ b/.github/actions/always/action.yml
@@ -9,6 +9,9 @@ inputs:
   secrets:
     description: "JSON stringified object containing all the secrets from the calling workflow"
     required: true
+  needs:
+    description: "JSON stringified object containing all the needs from the calling workflow"
+    required: false
 runs:
   using: "composite"
   steps:
@@ -17,8 +20,10 @@ runs:
       env:
         INPUTS: ${{ inputs.json }}
         SECRETS: ${{ inputs.secrets }}
+        NEEDS: ${{ inputs.needs }}
         GITHUB: ${{ toJSON(github) }}
       run: |
         echo "${INPUTS}"
         echo "${SECRETS}"
+        echo "${NEEDS}"
         echo "${GITHUB}"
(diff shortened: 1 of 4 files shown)


issenn:
Can `needs` be passed to action ?

Why can't uses the fork repository as a reusable workflow ?
Like this `uses: issenn/flowzone/.github/workflows/flowzone.yml@v0.54.0` will got an `workflow not found` error.

ab77:
> Can `needs` be passed to custom composite action ?

@klutchell do you recall why we didn't set `restrict_custom_actions: true` here already? For the other part, I don't see any issues passing the `needs` object to custom actions...

> Why can't uses the fork repository as a reusable workflow ? Like this `uses: issenn/flowzone/.github/workflows/flowzone.yml@v0.54.0` will got an `workflow not found` error.

@issenn that should work - did you check your repo/org settings to see if reusable workflows are enabled/unrestricted?

issenn:
> > Why can't uses the fork repository as a reusable workflow ? Like this `uses: issenn/flowzone/.github/workflows/flowzone.yml@v0.54.0` will got an `workflow not found` error.
> 
> @issenn that should work - did you check your repo/org settings to see if reusable workflows are enabled/unrestricted?

I can't find related options in repo settings, only see `Allow all actions and reusable workflows` enabled in `Settings`-`Action`-`General`-`Actions permissions`. I didn't change any settings, I'll try again later with fork repository.


> except adding a new input at this point would break any existing custom actions that aren't expecting it. Might be better to find a way to do it with env vars like we did for custom matrix, etc.

I think, adding a new input element in the action does not affect other actions, because this element is unused. `needs` has some pre-initialized `event_types` and `project_types`, and `env.matrix_value` can only pass matrix strategy parameters.

klutchell:
Hey @issenn , do you still need this change? Would you mind updating your PR to use the change I suggested above to avoid breaking existing workflows?
```
echo "needs_json=${{ toJSON(needs) }}" >> $GITHUB_ENV
```

PR #426, opened 2023-01-14
Restore case insensitive regex

klutchell:
Change-type: patch


klutchell:
approve this please

PR #428, opened 2023-01-16
Only download the required npm artifacts for finalize

klutchell:
The entire download step fails if other artifacts from the workflow run have expired. Make sure we only grab the npm artifacts with the long expiry time.

Change-type: patch


pipex (approved):
Smart solution, it should prevent the issue in the majority of cases. LGTM

PR #429, opened 2023-01-16
Allow continue on error when pushing versioned commit

klutchell:
If we need to re-run a merge job due to a failed publish, we don't want to be blocked by the fact that the commit already exists in the base branch.

Change-type: patch


klutchell:
i self-certify

PR #430, opened 2023-01-16
Index the first element of the node_versions array

klutchell:
The -1 syntax doesn't work and it shouldn't matter which version we are using here.

Change-type: patch


klutchell:
I self-certify!

PR #431, opened 2023-01-17
Update previous tag search to match the v prefix

klutchell:
We encountered cases in balena-engine where the tag sorting did not return the expected most recent v tag.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -1581,11 +1581,13 @@ jobs:
           # prevent git from existing with 141
           set +o pipefail
           previous_tag="$(git tag -l --sort=-version:refname "v*" | head -n1)"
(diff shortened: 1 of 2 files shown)


klutchell:
> @klutchell , can tags like `v22.06.0-beta.0` that sometimes we get from upstream (in the balenaEngine case) cause problems?
> 
> (I guess we should not pull tags from Moby, but we happen to have a `v22.06.0-beta.0` tag there -- I believe I can safely remove it if necessary.)

We pull all tags from Moby, that's the main issue here. I can try to filter further but this check is better than what we have today. Try running the old previous_tag command in your workspace and see what happens.

```
git tag --sort=-version:refname | head -n 2 | tail -n 1
```

We need to make sure to allow balena-os tags like `1.2.3-rev1` so I don't want to spend too much time on tag filters if we can avoid it.

ab77 (approved):

PR #435, opened 2023-01-18
Fix parsing of previous tag for release notes

klutchell:
Change-type: patch
See: https://github.com/product-os/flowzone/pull/431


klutchell:
I self-certify!

PR #436, opened 2023-01-18
Update renovate ignore paths to the defaults from config:base

klutchell:
Change-type; patch


klutchell:
I self-certify!

PR #438, opened 2023-01-20
Move workflow de-duplication condition into flowzone core

klutchell:
This check to prevent duplicate runs when both pull_request and pull_request_target are triggered is always required, and is better defined inside Flowzone itself rather than the calling workflow.

Change-type: minor


klutchell:
I self-certify!

PR #443, opened 2023-01-24
Add proper documentation for website builder

anujdeshpande:
Change-type: patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,9 @@ If you have an `npm run doc` script then it will automatically be run and the ge
 
 ### Website
 
-If you have docs that you intend to publish on a website, checkout the [Getting Started](https://docusaurus-builder.pages.dev/) section of the Docusaurus builder. The docs will be built using the Docusaurus framework and published on Cloudflare Pages. 
+1. Ask admin to create a new Cloudflare Pages Project for you and share the project name
+2. Add the project name to your `flowzone.yml` under the `cloudflare_website` input
+3. Admin will also set the deployment branch for you on the Cloudflare Dashboard (typically master or main - depending on what you chose when creating the repo)
 
 If you intend to use a custom framework for your docs build, then you can make use of the custom website build option by adding your desired build command in a input called `custom_website_build`. This command should generate your static site into a folder called `build` which will then be deployed to Cloudflare Pages. 
 


Bucknalla (commented):
@anujdeshpande worth being more clear that `cloudflare_website` gives you both the docusaurus and the cloudflare deployment. I do wonder if we should have `docusaurus` or something that builds the site and puts it on a randomly generated URL and then `cloudflare_website` that points it at your specified subdomain.

vipulgupta2048:
I think that simplification would result in extra friction being added to the process. I don't see any use for a randomly generated URL as compared to the name-based generated one. What it would do is add extra code for automating more of Cloudflare.

anujdeshpande:
@vipulgupta2048  @Bucknalla here's a thought - why would someone care about where it is hosted? Cloudflare should be an alien concept to someone who just wants a pretty looking docs site

Bucknalla:
Agreed, tbh the user shouldn't even need to know that it's cloudflare, they should just specify that they want a website and what subdomain it should have. Thus we abstract and hide all of that from the in a 'website builder' rather than having two slightly conflated workflows that overlap.

anujdeshpande:
@Bucknalla @vipulgupta2048 to add more clarity - I am now proposing that we rename the flowzone input for docs site to just `docusaurus_project` and then let ops fill that in flowzone.yml for anyone requesting a website. Because there are a couple of steps that ops has to do and just one step that the user has to do - let's just eliminate the user step, and keep it all for ops. All you do is say that you want a site and point to a repo that you want the site on.

Bucknalla:
What about removing the fact that it's docusaurus under the hood? `website: autokit-usermanual` for example. If we can error handle it better for when there isn't a cloudflare site set up, to suggest they should ask ops, that would be good but not necessary critical.

hraftery:
I'm not sure that we can remove the Docusaurus details. This was my big realisation from my docusaurus-101 session with @vipulgupta2048 a while back - as soon as `cloudflare_website` is added, (part of) the repo becomes docusaurus content that will be pre-processed by docusaurus-builder, processed by docusaurus and then published. So the repo's maintainers need to be familiar with (and have access to) the requirements of those tools.

So in my experience the user's tasks are:

1. Pick a project name and check it's not already taken <-- maybe in consultation if we want central coordination of these names - pending outcome of Zulip thread.
2. As ops to create it / add redirects.
3. Prepare the repo to suit - eg. take advantage of the separate READMEs for GitHub and for Docusaurus, add any frontmatter tags to suit our theme, update absolute/relative media URLs, consider the specialisations afforded by `docusaurus-builder`.
4. If the URL is public/known/linked, then preview locally.
5. Add maintainer documentation that explains that the repo contains docusaurus content, and how to work with it, referencing `docusaurus-builder`.
6. When ready, throw the switch and edit `flowzone.yml`.

We've already set the precedent that the repo maintainer is in charge of the `flowzone.yml` when we made the switch in the first place. We centrally administer Flowzone, but expect maintainers to manage their own `flowzone.yml`. I think it's a good precedent to maintain, since being aware of and having control over of your CI/CD is important when crafting PR's to suit.

So I'm in favour of keeping some user steps explicit.

As far as the file in this PR is concerned, I prefer the original content (but with a corrected URL!) since docs on using `docusaurus-builder` ought to live with `docusaurus-builder`, which also avoids redundancy. I expect the Flowzone docs just to say Flowzone supports it and alternatives, and to define each "input" (funny name, btw).

On the separation of cloudflare/docusaurus, I agree `cloudflare_website` is a poor name because Cloudflare should be a fungible implementation detail, unlike docusaurus which the user must design for. But again, perhaps slightly out of scope for this PR. I don't mind "Cloudflare" appearing in the docs here, since the detail is important to know and this is a reasonable place to mention it.

Issue #442, opened 2023-01-23
Get latest tag for current branch returns error


klutchell:
```
Run tag="$(git tag --list --sort=-version:refname "v*.*.*" --merged | head -n1)"
  tag="$(git tag --list --sort=-version:refname "v*.*.*" --merged | head -n1)"
  echo "semver=${tag/v/}" >> $GITHUB_OUTPUT
  echo "tag=${tag}" >> $GITHUB_OUTPUT
  shell: /usr/bin/bash --noprofile --norc -eo pipefail -x {0}
  env:
    GHCR_USER: flowzone
    NPM_REGISTRY: registry.npmjs.org
    CARGO_REGISTRY: crates.io
++ git tag --list --sort=-version:refname 'v*.*.*' --merged
++ head -n1
+ tag=v64.0.[2](https://github.com/balena-io/balena-api/actions/runs/3988212874/jobs/6839089676#step:8:2)
Error: Process completed with exit code 1[4](https://github.com/balena-io/balena-api/actions/runs/3988212874/jobs/6839089676#step:8:4)1.
```
This error is ignored and the build proceeds, but it should be cleaned up as it might be impacting repos with custom versioning.

klutchell:
Still an issue: https://github.com/balena-io/balena-api/actions/runs/4058114928/jobs/6984707970

PR #445, opened 2023-01-24
Disable pager for git log commands to avoid error 141

klutchell:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/442


klutchell:
I self-certify!

PR #446, opened 2023-01-24
Pre-process compose files so we can modify them before tests

klutchell:
Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -401,8 +401,7 @@ jobs:
       npm: ${{ steps.npm.outputs.enabled }}
       npm_private: ${{ steps.npm.outputs.private }} # can be null or unset
       npm_docs: ${{ steps.npm.outputs.docs }} # can be null or unset
-      docker_compose: ${{ steps.docker_compose.outputs.file }} # can be null or unset
-      docker_compose_test: ${{ steps.docker_compose_test.outputs.file }} # can be null or unset
+      docker_compose_yaml: ${{ steps.docker_compose.outputs.yaml }}
       docker_bake_json: ${{ steps.docker_bake.outputs.json }}
       docker_bake_matrix: ${{ steps.docker_bake.outputs.matrix }}
       balena: ${{ steps.balena.outputs.enabled }}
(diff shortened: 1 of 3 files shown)


klutchell:
I self-certify!

flowzone.yml
```diff
@@ -793,19 +770,40 @@ jobs:
...
+      - name: Check for Docker compose files
+        id: docker_compose_files
+        uses: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5 # v0.2.1
+        with:
+          cmd: bash -c "find -L ${{ github.workspace }}/${{ inputs.working_directory }} -maxdepth 1 -type f -name 'docker-compose*.yml' -o -name 'docker-compose*.yaml'" || true
+          separator: newline
+
+      # only if docker-compose test files are found
+      # merge all compose files in the project and record the merged config as a multiline yaml output
+      - name: Pre-process Docker compose files
```
ab77, line 782: @klutchell this steps breaks since it's missing .env file ([e.g.](https://github.com/balena-io/environment-staging/actions/runs/4001511798/jobs/6867816155#step:39:14))

Do we need to move/copy .env generation from the docker steps below up here or would `touch .env` suffice?

flowzone.yml
```diff
@@ -1249,22 +1248,25 @@ jobs:
           driver-opts: network=host
           install: true
 
+      # set env var TARGET that is just a sanitized matrix.target
+      # set env var PREFIX if target is not 'default'
+      # set env var PLATFORM that is just a sanitized matrix.platform
+      # write compose yaml to file and set platform={{matrix.platform}} on any services with the 'build' directive
+      # write docker bake json to file
       - name: Set env vars
```
ab77, line 1256: @klutchell also breaks [here](https://github.com/balena-io/open-balena-vpn/actions/runs/4000612249/jobs/6866051809#step:6:80).

klutchell: I just noticed that too and this fix is [here](https://github.com/product-os/flowzone/pull/448)

flowzone.yml
klutchell: I think I can just write our generate compose to the same working directory as the other compose files and we will be good. Let me open a PR and change the tests.

PR #447, opened 2023-01-24
Save image to file before running compose tests

klutchell:
This prevents accidentally uploading images that were built with docker-compose tests.

Change-type: patch


klutchell:
I self-certify!

PR #449, opened 2023-01-25
Restore target prefix for non-default targets

klutchell:
This bug prevented the target name from being
prefixed on docker images tags.

Change-type: patch


klutchell:
I self-certify!

PR #451, opened 2023-01-25
Saves the COMPOSE_VARS secret to .env for docker-compose pre-processing.

cmfcruz:
This fixes the `no such file or directory` error when docker compose pre-processing (`docker compose config`) is done for a docker-compose that requires a `.env` file.

Change-type: patch
Signed-off-by: Carlo Miguel F. Cruz <carloc@balena.io>


cmfcruz:
I self-certify!

Issue #456, opened 2023-01-26
Use `-f` flag for multi docker-compose merge instead of COMPOSE_FILE


cywang117:
After checking out issue https://github.com/product-os/flowzone/issues/454 and PR https://github.com/product-os/flowzone/pull/455, I found that merging multiple files with `COMPOSE_FILE=file-one.yml:file-two.yml` doesn't work as expected. Example:

`docker-compose.yml`:
```
version: '3'
services:
  main:
    image: alpine
    command: sh -c "echo main && sleep infinity"
    network_mode: host
```
`docker-compose.test.yml`
```
version: '3'
services:
  main:
    image: alpine
    command: sh -c "echo test && sleep infinity"
    network_mode: "bridge"
    networks:
      - default
```
Running `COMPOSE_FILE="$(echo $(ls -1 docker-compose.{yml,yaml} docker-compose.test.{yml,yaml} 2>/dev/null) | sed 's/ /:/')" docker compose config` and also when the file order is reversed both return:
```
services:
  main:
    command:
    - sh
    - -c
    - echo main && sleep infinity
    image: alpine
    network_mode: host
    networks:
      default: null
networks:
  default:
    name: test_default
```
However, running `docker-compose -f docker-compose.yml -f docker-compose.test.yml config` with the above files does work:
```
services:
  main:
    command:
    - sh
    - -c
    - echo test && sleep infinity
    image: alpine
    network_mode: bridge
    networks:
      default: null
networks:
  default:
    name: test_default
```

Therefore, switching to `-f` will support test env overrides the most.

Issue #454, opened 2023-01-26
Preprocess docker-compose files so that `docker-compose.test.yml` fields take precedence


cywang117:
Currently, when pre-processing docker-compose files for the "Test docker" step, `docker-compose.yml` fields take precedence over `docker-compose.test.yml`: 

https://github.com/product-os/flowzone/blob/fcc8f01f3619b577ccb7ac118a0e2b8bf0968374/flowzone.yml#L773-L798

Given that the preprocessing is done for a test environment, `docker-compose.test.yml` fields should take precedence.

PR #457, opened 2023-01-26
Use -f for multi-docker-compose instead of COMPOSE_FILE

cywang117:
Closes: #456
See: #454
Change-type: patch
Signed-off-by: Christina Ying Wang <christina@balena.io>


klutchell:
lgtm!

PR #458, opened 2023-01-26
Add balena release notes

ab77:
Set git release notes on balena releases (e.g.)
https://dashboard.balena-cloud.com/apps/1957201/releases/2455644

image

Could make it more useful in the future and add a portion of the change log instead of `git tag..tag ..` output.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -88,6 +88,7 @@
 
   - &deployToBalenaAction
     uses: balena-io/deploy-to-balena-action@master
+    id: balena_deploy
     with:
       balena_token: ${{ secrets.BALENA_API_KEY || secrets.BALENA_API_KEY_PUSH }}
       environment: ${{ inputs.balena_environment }}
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -44,6 +45,42 @@
...
+  # FIXME: https://github.com/balena-io/deploy-to-balena-action/issues/194
+  - &updateBalenaReleaseNotes
+    name: Update balena release notes
+    run: |
+      set -ea
+      # prevent git from existing with 141
+      set +o pipefail
+
+      previous_tag="$(git --no-pager tag --list --sort=-version:refname "v*.*.*" --merged | head -n2 | tail -n1)"
+      release_notes="$(git log ${previous_tag}..${{ github.event.pull_request.head.sha || github.event.head_commit.id }} --pretty=reference | jq -R -s '.')"
```
klutchell, line 57: At this point it might be worth templating a new `&getReleaseNotes` step since we are doing it more than once?

The output being a format that could be consumed by both balena deploy and github release jobs?

flowzone.yml
```diff
@@ -44,6 +45,42 @@
...
+
+      app_id="$(curl "https://api.${{ inputs.balena_environment }}/v6/application?\$filter=slug%20eq%20%27${{ matrix.slug }}%27&\$select=id" \
+        -H 'Content-Type: application/json' \
+        -H 'Authorization: Bearer ${{ secrets.BALENA_API_KEY || secrets.BALENA_API_KEY_PUSH }}' \
+        | jq -r '.d[].id')"
+
+      release_id='${{ steps.balena_deploy.outputs.release_id }}'
+      # FIXME: https://github.com/balena-io/deploy-to-balena-action/issues/193
```
klutchell, line 65: We should see if someone is willing to fix this, myself included, before committing this workaround IMO.

Maybe it's a 1min fix but no one has looked yet. I'll add it to my list.

flowzone.yml
```diff
@@ -44,6 +45,42 @@
...
+            "https://api.${{ inputs.balena_environment }}/v6/release?\$filter=belongs_to__application%20eq%20${app_id}&\$select=id,semver,is_final,is_invalidated,release_type" \
+            -H 'Content-Type: application/json' \
+            -H 'Authorization: Bearer ${{ secrets.BALENA_API_KEY || secrets.BALENA_API_KEY_PUSH }}' \
+            | jq -r --arg semver '${{ needs.versioning.outputs.semver }}' \
+            '.d[] | select((.semver==$semver) and (.is_final==true) and (.is_invalidated==false) and (.release_type=="final")).id' \
+            | head -n 1)
+      fi
+
+      if [[ -n $release_notes ]] && [[ -n $app_id ]] && [[ -n $release_id ]]; then
+          curl --retry 3 --fail \
```
klutchell, line 77: We should suppress the progress of curl commands in these scripts.

flowzone.yml
ab77: Yeah, that's not a bad idea actually.

flowzone.yml
klutchell: See https://github.com/balena-io/deploy-to-balena-action/pull/246

klutchell:
lgtm

PR #461, opened 2023-01-30
Encode docker compose config as base64 to preserve special characters

klutchell:
Change-type: patch

In some cases the formatting of a compose file is broken when exported as an environment variable in GitHub actions. See this example [failed job](https://github.com/klutchell/unbound-docker/actions/runs/4043638837/jobs/6952826663).


klutchell:
I self-certify!

PR #462, opened 2023-01-31
Open draft PRs in downstream repos for each Flowzone PR

klutchell:
Change-type: minor
See: https://github.com/product-os/flowzone/issues/63

This doesn't actually validate the steps of the downstream jobs, just tries to identify if any failures were introduced by the Flowzone PR in question.

If downstream repos are unstable they should be removed from the self-checks and replaced with another repo.


klutchell:
I self-certify!

PR #468, opened 2023-02-01
Make the Rust toolchain version user configurable

pipex:
Adds a `rust_toolchain` (defaults to `stable`) to allow users to configure their preferred toolchain to use when building Rust binaries.

Change-type: minor

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -149,6 +149,11 @@ on:
           arm-unknown-linux-gnueabihf,
           x86_64-unknown-linux-gnu,
           i686-unknown-linux-gnu
+      rust_toolchain:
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -1978,8 +1982,9 @@ jobs:
       - *extractSourceArtifact
 
       - name: Set up toolchain ${{ matrix.target }}
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@master
```
klutchell, line 1985: I don't think you can use variables in the `uses` field?

pipex:
I self-certify!

PR #472, opened 2023-02-01
Renove balena-cli from e2e tests and re-enable fatrw

klutchell:
Change-type: patch


klutchell:
I self-certify!

PR #480, opened 2023-02-03
Restore old tag lookup behaviour

klutchell:
If versioning is disabled and a PR is open, there
will likely not be a tag on the HEAD commit.

However we still want to use something to tag
our artifacts so use git descibe.

Change-type: patch


klutchell:
I self-certify!

Issue #419, opened 2023-01-11
Do not cancel concurrent merge events


klutchell:
We automatically cancel concurrent Flowzone events with the same GitHub ref, but we should never cancel merge events as they are applying the versioned source to master.

https://github.com/product-os/flowzone/blob/a5b8dfce0d2c0666f09b2a753355912d364a1cff/flowzone.yml#L298-L301

Ideally for PR open/sync events we cancel existing runs, but for merge events we run them in series, if possible.

https://docs.github.com/en/actions/using-jobs/using-concurrency

PR #486, opened 2023-02-03
Cancel concurrent jobs with the exception of merged PRs

klutchell:
We always want merged PRs to queue up and complete in order or versioning will break.

Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/419


klutchell:
I self-certify!

Issue #193, opened 2022-09-08
set access=restricted when publishing npm from private repos


klutchell:
The resinCI flow is here:
- https://github.com/product-os/scripts/blob/652fdd41851cda614a6a4c29cf7a14ce17bc9b6b/npm/store.sh#L9-L17
- https://github.com/product-os/resinci-deploy/blob/9af1ae97baaf94906ebcbf80531ce5313d9e19c8/lib/npm.ts#L39-L67

klutchell:
Apparently this flag does nothing on existing npm repos
https://docs.npmjs.com/cli/v8/commands/npm-publish#access

> Note: Using the --access flag on the npm publish command will only set the package access level on the initial publish of the package. Any subsequent npm publish commands using the --access flag will not have an effect to the access level. To make changes to the access level after the initial publish use npm access.

PR #487, opened 2023-02-03
Set npm publish access based on repo visibility

klutchell:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/193


flowzone.yml
```diff
@@ -1309,7 +1326,7 @@ jobs:
           if [ ${{ github.run_attempt }} -gt  1 ]; then
             npm --loglevel=verbose --logs-max=0 unpublish ${{ needs.npm_test.outputs.package }}@${{ needs.npm_test.outputs.version_tag }}-$((${{ github.run_attempt }} - 1)) || true
           fi
-          npm --loglevel=verbose --logs-max=0 publish --tag=${{ needs.npm_test.outputs.sha_tag }} "${pack}"
+          npm --loglevel=verbose --logs-max=0 publish --tag=${{ needs.npm_test.outputs.sha_tag }} "${pack}" --access="${{ needs.npm_check.outputs.access }}"
```
klutchell, line 1329: If the GH repo is public, should we just omit the access flag altogether? If the flag is not set at publish time, does it inherit access settings from the npm repo?

Page-: If it's not set it looks like it'll either be "restricted" if it's a scoped package or public if it's unscoped, and setting it here will change the existing setting (useful for going private -> public or public -> private, although going public -> private will obviously not be fully guaranteed to cover everything)

klutchell: Apparently this flag does nothing on existing npm repos
https://docs.npmjs.com/cli/v8/commands/npm-publish#access

> Note: Using the --access flag on the npm publish command will only set the package access level on the initial publish of the package. Any subsequent npm publish commands using the --access flag will not have an effect to the access level. To make changes to the access level after the initial publish use npm access.

klutchell:
I self-certify!

PR #493, opened 2023-02-06
Allow pulling private base images for private repositories

klutchell:


Page- (approved):

build.js
```diff
@@ -5,7 +5,7 @@ const yaml = require('yaml');
 // this means that yaml anchors will be "exploded" and not present in the github workflow
 // we write out, which means we can use anchors when developing flowzone but not have to
 // worry about github actions not supporting them
-const flowzone = yaml.parse(fs.readFileSync('./flowzone.yml', 'utf8'));
+const flowzone = yaml.parse(fs.readFileSync('./flowzone.yml', 'utf8'),{merge: true});
```
Page-, line 8: The correctly formatted version (if we had balena-lint set up on this repo) would be:
```suggestion
const flowzone = yaml.parse(fs.readFileSync('./flowzone.yml', 'utf8'), {
	merge: true
});
```

flowzone.yml
```diff
@@ -53,6 +53,41 @@
...
+  - &ifInternalPullRequest
+    if: github.event.pull_request.head.repo.full_name == github.repository
+
+  - &ifPrivateRepository
+    if: github.event.repository.private
+
+  - &loginWithDockerHub
+    name: Login to Docker Hub
+    continue-on-error: true
+    uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9 # v1
```
Page-, line 65: This updates from `f4ef78c080cd8ba55a85445d5b36e214a81df20a` for a bunch of the uses - is that fine/safe?

klutchell: Good catch, they should be at v2, this is actually an accidental downgrade.

flowzone.yml
```diff
@@ -53,6 +53,41 @@
...
+      username: ${{ secrets.DOCKERHUB_USER || secrets.DOCKER_REGISTRY_USER }}
+      password: ${{ secrets.DOCKERHUB_TOKEN || secrets.DOCKER_REGISTRY_PASS }}
+
+  - &loginWithGitHubContainerRegistry
+    name: Login to GitHub Container Registry
+    continue-on-error: true
+    uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2
+    with:
+      registry: ghcr.io
+      username: ${{ github.actor }}
```
Page-, line 76: This changes a bunch of entries from `env.GHCR_USER` to `github.actor` - is that ok?

klutchell: ugh, no

Page- (approved):

PR #494, opened 2023-02-07
Don't leak compose secrets

ab77:
Prevent leaking compose secrets during masking due to `-x` shell attribute being set in the `defaults` section.


ab77:
lgtm

PR #495, opened 2023-02-07
Avoid leaking secrets from COMPOSE_VARS

klutchell:
Change-type: patch


klutchell:
I self-certify!

PR #496, opened 2023-02-07
Use self-hosted runners for Docker builds

klutchell:


klutchell:
Need to make sure we only apply self-hosted tags for balena enterprise orgs/repos.

klutchell:
We can use the GH API to check how many runners are available at the org level, and the repo level.

If either of those checks fail due to insufficient permissions we will handle the failure and treat that as 0 runners.

If 1 or more runners are available (maybe matching our preferred labels) we will attempt to use them. Otherwise keep using GH hosted runners.

klutchell:
I self-certify!

PR #503, opened 2023-02-09
Ephemeral access tokens

ab77:
@klutchell this PR should be ready to go, please review when you get a chance. There are some structural changes in here, especially around tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,21 +65,51 @@ jobs:
         # update the branch protection rules on the Flowzone repository settings
         repository:
           [
+            balena-io-modules/open-balena-base,
             balena-io/balena-api,
             balena-io/docs,
             balena-io/environment-staging,
-            balena-io-modules/open-balena-base,
             balena-os/balena-engine,
+            balena-os/fatrw,
             balenaltd/handbook,
             product-os/environment-staging,
-            balena-os/fatrw,
           ]
+        include:
+          - token_scope: >-
+              {
+                "actions": "read",
+                "administration": "write",
+                "checks": "read",
+                "contents": "write",
+                "members": "read",
+                "metadata": "read",
+                "pull_requests": "write",
+                "statuses": "read",
+                "workflows": "write"
+              }
+          - app_id: "${{ vars.APP_ID || '291899' }}"
+          - installation_id: 34046903
+            repository: balena-io-modules/open-balena-base
+          - installation_id: 34046749
+            repository: balena-io/balena-api
+          - installation_id: 34046749
+            repository: balena-io/docs
+          - installation_id: 34046749
+            repository: balena-io/environment-staging
+          - installation_id: 34046907
+            repository: balena-os/balena-engine
+          - installation_id: 34046907
+            repository: balena-os/fatrw
+          - installation_id: 34046911
+            repository: balenaltd/handbook
+          # (e.g. https://github.com/organizations/product-os/settings/installations/34040165)
+          - installation_id: 34040165
+            repository: product-os/environment-staging
 
     env:
       # environment variables used by gh CLI
       # https://cli.github.com/manual/gh_help_environment
       GH_REPO: ${{ matrix.repository }}
-      GH_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
       GH_PROMPT_DISABLED: "true"
       GH_DEBUG: "true"
       GH_PAGER: "cat"
@@ -87,9 +117,37 @@ jobs:
       pr_title: "Test Flowzone PR ${{ github.event.pull_request.number }}"
 
     steps:
+      - name: Check secrets
+        id: private_key
+        run: |
+          if [[ -z '${{ secrets.FLOWZONE_TOKEN }}' ]] \
+            && [[ -z '${{ secrets.GH_APP_PRIVATE_KEY }}' ]]; then
+              echo '::error::Must specify either GH_APP_PRIVATE_KEY or FLOWZONE_TOKEN.'
+              false
+          fi
+
+          if [[ -n '${{ secrets.GH_APP_PRIVATE_KEY }}' ]]; then
+              echo 'available=true' >> $GITHUB_OUTPUT
+          else
+              echo 'available=false' >> $GITHUB_OUTPUT
+          fi
+
+      - name: Generate token
+        uses: tibdex/github-app-token@v1.8.0
+        if: steps.private_key.outputs.available == 'true'
+        id: gh_token
+        with:
+          app_id: ${{ matrix.app_id }}
+          installation_id: ${{ matrix.installation_id }}
+          permissions: ${{ matrix.token_scope }}
+          private_key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+          repository: ${{ matrix.repository }}
+
       - name: Get repo settings
         if: github.event.action != 'closed'
         id: repo
+        env:
+          GH_TOKEN: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
         run: |
           echo "default_branch=$(gh api repos/{owner}/{repo} --jq '.default_branch')" >> $GITHUB_OUTPUT
 
@@ -97,7 +155,7 @@ jobs:
         if: github.event.action != 'closed'
         with:
           repository: ${{ matrix.repository }}
-          token: ${{ secrets.FLOWZONE_TOKEN }}
+          token: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
           ref: ${{ steps.repo.outputs.default_branch }}
 
       # import gpg key for signed commits
@@ -122,6 +180,7 @@ jobs:
           GIT_AUTHOR_EMAIL: ${{ steps.import-gpg.outputs.email }}
           GIT_COMMITTER_NAME: ${{ steps.import-gpg.outputs.name }}
           GIT_COMMITTER_EMAIL: ${{ steps.import-gpg.outputs.email }}
+          GH_TOKEN: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
         run: |
           yq '(.jobs.* | select(.uses | contains("flowzone.yml"))).uses |= sub("(?P<uses>.+)@.+"; "$1@${{ github.event.pull_request.head.sha }}")' -i .github/workflows/flowzone.yml
 
@@ -138,7 +197,7 @@ jobs:
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
         id: cpr
         with:
-          token: ${{ secrets.FLOWZONE_TOKEN }}
+          token: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
           branch: ${{ env.pr_branch }}
           title: "${{ env.pr_title }}"
           body: |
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -79,25 +79,57 @@ jobs:
...
+      INSTALLATION_ID: "${{ vars.INSTALLATION_ID || 34040165 }}"
+      TOKEN_SCOPE: >-
+        {
+          "administration": "write"
+        }
       pr_branch: "flowzone/${{ github.head_ref || github.ref_name }}"
       pr_title: "Test Flowzone PR ${{ github.event.pull_request.number }}"
 
     steps:
+      - name: Check secrets
```
ab77, line 95: unfortunately we can't output the token from previous job, since it's masked, so have to duplicate code here.

.github/workflows/tests.yml
```diff
@@ -151,52 +210,41 @@ jobs:
...
+          filter_re: "^policy-?bot"
+          GH_TOKEN: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
         run: |
-          while true
-          do
-            sleep 30
-
-            prev_checks="${checks}"
-
-            checks="$(gh pr status --json statusCheckRollup,number --jq '[.createdBy[] |
```
ab77, line 172: `.createdBy` is empty when queried using ephemeral tokens. Must be a GH permissions thing requiring a token tied to a user (not machine/app) identity. No way around it as far as I can tell. I've re-implemented using `gh pr checks`, which is a bit kludgy but seems to work).

klutchell:
@ab77 I hope to review this soon, I just have to wrap up some other tasks!

README.md
```diff
@@ -330,11 +334,17 @@ but they can also be [configured for personal repositories](https://docs.github.
 
 These secrets can also be found at the top of [flowzone.yml](./.github/workflows/flowzone.yml).
 
+#### `PRIVATE_KEY`
```
Page-, line 337: I'd prefer a more specific name than `PRIVATE_KEY`, eg `GH_APP_PRIVATE_KEY`
```suggestion
#### `GH_APP_PRIVATE_KEY`
```
or maybe adding the ephemeral bit in there as well could make sense?

flowzone.yml
```diff
@@ -109,11 +119,15 @@
     run: gh release delete --yes '${{ github.event.pull_request.head.ref }}' || true
     env:
       <<: *gitHubCliEnvironment
+      GH_TOKEN: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
```
Page-, line 122: This is repeated a lot, it could make sense to use anchors/aliases, eg
```suggestion
      <<:
        - *gitHubCliEnvironment
        - *ghTokenEnv
```
```yaml
- &ghTokenEnv
  GH_TOKEN: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}

Comment (klutchell) [reply flowzone.yml:122]: +1

File: .github/workflows/tests.yml [hunk 5]
```diff
@@ -65,39 +65,96 @@ jobs:
...
+          fi
+
+          if [[ -n '${{ secrets.PRIVATE_KEY }}' ]]; then
+              echo 'available=true' >> $GITHUB_OUTPUT
+          else
+              echo 'available=false' >> $GITHUB_OUTPUT
+          fi
+
+      - name: Generate token
+        uses: tibdex/github-app-token@v1
```
klutchell, line 136: We should pin this to a full semver since it's not an official github action.

Once this merges Renovate will pin it to a commit hash, and keep the version as a comment for tracking.

flowzone.yml
```diff
@@ -41,7 +54,6 @@
     with:
       fetch-depth: 0
       submodules: "recursive"
-      token: ${{ secrets.FLOWZONE_TOKEN }}
```
klutchell, line 44: Why were these removed? If the intent is to use `${{ secrets.GITHUB_TOKEN }}` instead I would rather be clear about that rather than appear to have forgotten to specify.

flowzone.yml
```diff
@@ -75,10 +86,10 @@
     with:
       registry: ghcr.io
       username: ${{ env.GHCR_USER || github.actor }}
-      password: ${{ secrets.GHCR_TOKEN || secrets.FLOWZONE_TOKEN }}
+      password: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
```
klutchell, line 89: If we are deprecating this secret we need to be clear about that. I thought it was preferable to allow the option for a different token with package write access. Maybe that's not required, but we should be clear that the secret is unused going forward.

flowzone.yml
```diff
@@ -92,7 +103,6 @@
   - &gitHubCliEnvironment
     # environment variables used by gh CLI
     # https://cli.github.com/manual/gh_help_environment
-    GH_TOKEN: "${{ secrets.FLOWZONE_TOKEN }}"
```
klutchell, line 95: I think it would make more sense to provide the token here
```
GH_TOKEN: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
```
and just ensure we only call this anchor from the step context, and not the job. I think the only place we set it at the job context is protect branch anyway.

That avoids having to have two aliases whenever we need to set gh env.

flowzone.yml
klutchell: or as I suggested above, keep this line in `*gitHubCliEnvironment`

flowzone.yml
```diff
@@ -188,7 +206,45 @@ on:
...
         required: false
+
     inputs:
+      app_id:
+        description: "GitHub App id to impersonate"
+        type: string
+        required: false
+        # https://github.com/organizations/product-os/settings/apps/flowzone-app
+        # https://github.com/organizations/product-os/settings/variables/actions/APP_ID
+        default: "${{ vars.APP_ID || 291899 }}"
```
klutchell, line 217:
```suggestion
        default: "${{ vars.APP_ID || '291899' }}"
```

flowzone.yml
```diff
@@ -188,7 +206,45 @@ on:
...
+      # not needed if installed on this current org/repo
+      installation_id:
+        description: "GitHub App installation id"
+        type: string
+        required: false
+        # https://github.com/organizations/product-os/settings/installations
+        # https://github.com/organizations/product-os/settings/variables/actions/INSTALLATION_ID
+        default: "${{ vars.INSTALLATION_ID || 34040165 }}"
```
klutchell, line 225:
```suggestion
        default: "${{ vars.INSTALLATION_ID || '34040165' }}"
```

flowzone.yml
```diff
@@ -188,7 +206,45 @@ on:
...
+            "members": "read",
+            "metadata": "read",
+            "organization_secrets": "read",
+            "packages": "write",
+            "pages": "write",
+            "pull_requests": "read",
+            "secrets": "read",
+            "statuses": "read",
+            "workflows": "read"
+          }
```
klutchell, line 247: Could we require this to be base64 encoded so it's less likely to get mangled in a workflow file?

flowzone.yml
```diff
@@ -457,6 +513,22 @@ jobs:
     steps:
       - *logGitHubContext
 
+      # https://stackoverflow.com/a/70249520/1559300
+      - name: Check secrets
```
klutchell, line 517: I think this got moved to the wrong job during a rebase. I would expect this step to be part of sanitize inputs.

.github/workflows/tests.yml
klutchell: I like how much cleaner it is, but I think `gh pr checks --required` will never pass if we change the name of a job.

These draft PRs intentionally don't update branch protections, so renamed jobs will be pending forever.

ab77: @klutchell you'll notice that we always filter the output and don't use the exit code from GH CLI.

flowzone.yml
ab77: @klutchell there is only place where `gitHubCliEnvironment` is used at the job/env level, so `steps` won't work there.

flowzone.yml
ab77: That was the intention.

flowzone.yml
ab77: I don't think there is a goods reason to keep it, I'll clean up the docs.

flowzone.yml
ab77: See, https://github.com/product-os/flowzone/blob/master/flowzone.yml#L2531-L2533

.github/workflows/tests.yml
```diff
@@ -151,52 +210,41 @@ jobs:
...
-            jq -r '.[].status' <<< "${checks}" | grep -q '^PENDING$' && continue
-
-            if [ "${prev_checks}" == "${checks}" ]
-            then
-              [ $(jq 'length' <<< "${checks}") -gt 0 ] || exit 1
-              jq -r '.[].status' <<< "${checks}" | grep -q '^COMPLETED$' && exit 0
-            fi
+          while ! gh pr checks --required \
+            --repo ${{ matrix.repository }} \
+            ${{ steps.cpr.outputs.pull-request-number}} | grep -qE "${flowzone_re}"; do
```
klutchell, line 223: Is it okay if one Flowzone check has loaded, but not the others? I think visually I have seen some required checks take longer to show up than others, and I wonder if we wont end up in another race condition with GitHub.

Maybe I'm just paranoid from being burned too many times.

flowzone.yml
ab77: @klutchell I would prefer this to be left as-is so it's clear what the token scope is. Looking at base64 strings isn't the best UX.

flowzone.yml
klutchell: In my other PR I removed all permissions from this token at the workflow level. So this will break on private repos since we removed read access.

Fortunately you can just set new permissions at the versioned_source and tagged_source jobs to get what you need. Or change the default at the top of the workflow and keep docker_test with empty permissions.

flowzone.yml
klutchell: Yeah, just move that to the steps instead of the job, so you can use a single alias.

flowzone.yml
klutchell: fair enough

.github/workflows/tests.yml
ab77: @klutchell let's catch up on this - it's a wider issue I think than just this PR. You are right, checks registration lags.

.github/workflows/tests.yml
```diff
@@ -151,52 +210,41 @@ jobs:
...
+              sleep $(((RANDOM % 5) + 5))
+          done
+          sleep $(((RANDOM % 15) + 15))
 
-            continue
+          while gh pr checks --required \
+            --repo ${{ matrix.repository }} \
+            ${{ steps.cpr.outputs.pull-request-number}} \
+            | grep -vE "${filter_re}" \
+            | grep -E '\s+pending\s+'; do
```
klutchell, line 233: I think ~~if~~ when we change the name of a required job we will get stuck here.

Because you are filtering on required, and looping if they are still pending. These PRs don't update the required checks to match the new names, so a renamed/refactored job will be missing and pending forever.

klutchell: Maybe if that's the case we can force merge the Flowzone PR, but it will come up at some point.

README.md
```diff
@@ -25,11 +25,11 @@ Reusable, opinionated, zero-conf workflows for GitHub actions
   - [Website](#Website)
 - [Customization](#customization)
   - [Secrets](#secrets)
+    - [`GH_APP_PRIVATE_KEY`](#gh_app_private_key)
     - [`FLOWZONE_TOKEN`](#flowzone_token)
     - [`GPG_PRIVATE_KEY`](#gpg_private_key)
     - [`GPG_PASSPHRASE`](#gpg_passphrase)
     - [`NPM_TOKEN`](#npm_token)
-    - [`GHCR_TOKEN`](#ghcr_token)
```
klutchell, line 32: I assume no one is using this in production anyway 👍

ab77: Yeah, and if someone has, it's a small change to refactor their workflows.

.github/workflows/tests.yml
ab77: .. and we also have these regexes in `parse_prepare_protection_rules`. Probably can leave it as is for now and if it ever comes up, we can look at parametrising the names?

klutchell: SGTM, maybe open a GH issue to investigate later

ab77:
lgtm

klutchell:
🎉

PR #505, opened 2023-02-13
Remove versioned source from protect branch steps

klutchell:
Also move check for draft PR to an anchor, and remove redundant dependencies.

Change-type: patch


klutchell:
I self-certify!

PR #509, opened 2023-02-15
Expose a github token with no permissions to docker build

klutchell:


flowzone.yml
```diff
@@ -1513,6 +1518,11 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.docker_check.outputs.docker_bake_matrix) }}
 
+    # remove all permissions from the default GITHUB_TOKEN
+    # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions
+    permissions: {}
```
klutchell, line 1524: duplicated permission logic at the job level in case we restore some permissions at the workflow level

just an extra safeguard against leaking secrets with actual permissions

ab77 (approved):
@klutchell LGTM, ~~however I would recommend merging ephemeral tokens PR first, since the rebase would be easier.~~ it's merged, not a big deal to rebase.

PR #510, opened 2023-02-16
Correct external contribution restrictions

klutchell:
Change-type: patch


klutchell:
I self-certify!

ab77 (approved):
@klutchell looks reasonable at a glance. At some point I think it would be worth instrumenting an external contribution test into the test workflow, somehow.

klutchell:
> @klutchell looks reasonable at a glance. At some point I think it would be worth instrumenting an external contribution test into the test workflow, somehow.

https://github.com/product-os/flowzone/issues/515

PR #516, opened 2023-02-17
Add missing ephemeral token steps

klutchell:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -1577,7 +1577,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ steps.gh_token.outputs.token || secrets.FLOWZONE_TOKEN }}
+          password: ${{ secrets.FLOWZONE_TOKEN || steps.gh_token.outputs.token }}
         if: github.event.repository.private
       - name: Login to Docker Hub
         continue-on-error: true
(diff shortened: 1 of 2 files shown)


klutchell:
@ab77 I think publishing to ghcr.io is broken until we figure out which permissions are missing from our ephemeral token.

For the Flowzone package I went to [package settings](https://github.com/orgs/product-os/packages/container/flowzone/settings) and under `Manage Actions access` I made sure that the Flowzone repo had write access. If this isn't the default we might need to enforce it via Flowzone.

ab77 (approved):

ab77 (approved):

klutchell:
@ab77 the existing app/token combination does not have permission to push a container to ghcr, I tried a number of combinations of permissions.

I also tried deleting the Flowzone package to see if it could be re-created with the ephemeral token, it cannot.

ab77:
> @ab77 the existing app/token combination does not have permission to push a container to ghcr, I tried a number of combinations of permissions.
> 
> I also tried deleting the Flowzone package to see if it could be re-created with the ephemeral token, it cannot.

I think this is the answer:

```
Error: failed to copy index: POST https://ghcr.io/v2/product-os/flowzone/blobs/uploads/: DENIED: permission_denied: installation not allowed to Create organization package
```

So GH App(s) (i.e. installations) not allowed to do this period. If you look [here](https://github.com/organizations/product-os/settings/apps/flowzone-app/permissions), there isn't a `packages` permission under the org. level, only repo.

But at lepo level, we have read/write:
image

klutchell:
@ab77 yup, but it was also failing to push to an existing package, 403, and we really need to be able to do both unless we want to involve ops for every new package and only allow pushing to ghcr for internal PRs.

PR #517, opened 2023-02-20
Inject github tokens to docker build and test

klutchell:
For external PRs just use the automatic github actions token
as the scope is already restricted.

Change-type: patch


klutchell:
What this PR needs is some aliases to resolve trusted vs untrusted tokens.

GetTokenForUntrusted would return either a generated app token or FLOWZONE_TOKEN for internal PRs, and the automatic GITHUB_TOKEN for external PRs.

GetTokenForTrusted would return either a generated app token or FLOWZONE_TOKEN for both internal and external PRs.

I'll add these on my next pass and install them where appropriate.

klutchell:
I self-certify!

PR #521, opened 2023-02-21
Use docker secrets to inject GitHub tokens

klutchell:
Allow Docker projects to use secrets to access
GITHUB_TOKEN and GH_TOKEN via Dockerfiles.

Change-type: patch


ab77 (approved):

klutchell:
Unclear if it's actually worthwhile to use secrets for compose tests. It might break existing compose files.

klutchell:
Marking as draft to re-evaluate the implementation of compose secrets as
there doesn't seem to be much benefit.

PR #523, opened 2023-02-22
Require 3 passing loops of required checks for draft PRs

klutchell:
This should avoid the race conditions where some required checks haven't populated yet, without adding significant time to drafts that are actually complete.

Change-type: patch


klutchell:
I self-certify!

Issue #518, opened 2023-02-20
Compress `docker save` images before uploading


Page-:
This will reduce the size used for build artifacts

PR #524, opened 2023-02-22
Compress docker artifacts before uploading

klutchell:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/518


klutchell:
I self-certify!

PR #528, opened 2023-02-27
Pass vars context to custom*

ab77:
Blocking: https://github.com/balena-io/remote-builders/pull/9

Any concern about custom actions affected in other repos can be easily fixed by pinning flowzone version in workflow.yml, or by updating the custom actions to include the new input.

### DIFF
--- a/.github/actions/always/action.yml
+++ b/.github/actions/always/action.yml
@@ -9,16 +9,21 @@ inputs:
   secrets:
     description: "JSON stringified object containing all the secrets from the calling workflow"
     required: true
+  variables:
+    description: "JSON stringified object containing all the variables from the calling workflow"
+    required: true
 runs:
   using: "composite"
   steps:
     - name: Dump context
       shell: bash
       env:
+        GITHUB: ${{ toJSON(github) }}
         INPUTS: ${{ inputs.json }}
         SECRETS: ${{ inputs.secrets }}
-        GITHUB: ${{ toJSON(github) }}
+        VARIABLES: ${{ toJSON(inputs.variables) }}
       run: |
+        echo "${GITHUB}"
         echo "${INPUTS}"
         echo "${SECRETS}"
-        echo "${GITHUB}"
+        echo "${VARIABLES}"
--- a/.github/actions/clean/action.yml
+++ b/.github/actions/clean/action.yml
@@ -9,16 +9,21 @@ inputs:
   secrets:
     description: "JSON stringified object containing all the secrets from the calling workflow"
     required: true
+  variables:
+    description: "JSON stringified object containing all the variables from the calling workflow"
+    required: true
 runs:
   using: "composite"
   steps:
     - name: Dump context
       shell: bash
       env:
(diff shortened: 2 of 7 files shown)


klutchell (dismissed):
What problem is this solving?

Adding a new input to custom actions will break every existing custom action being used in production, and afaik environment variables are already available to custom actions.

ab77:
> What problem is this solving?
> 
> Adding a new input to custom actions will break every existing custom action being used in production, and afaik environment variables are already available to custom actions.

We do need to support GitHub `vars` context for environments/deployments (e.g.):
https://github.com/balena-io/remote-builders/pull/9

.. because not everything is a secret and some values are useful to be configurable at runtime without hacking main/PRs, etc.

If stuff breaks as a result, so be it - it's an easy fix. We can't hold back features, just because someone's something breaks as a result, though I should make this a major or minor bump in that case.

klutchell:
But there are other ways to expose new fields, like encode it as JSON and write it to the environment.

ab77:
> But there are other ways to expose new fields, like encode it as JSON and write it to the environment.

Sure, there are almost infinite way of doing things, but what issue(s) specifically do you have with passing an existing GitHub object `vars` to custom actions as an input like this:

```
        with:
          variables: ${{ toJSON(vars) }}
```

... and consuming it int he custom composite actions like this:

```
---
name: always clean
inputs:
  variables:
    description: "JSON stringified object containing all the variables from the calling workflow"
    required: true
    ...

    - name: destroy test device
      shell: bash
      run: |
        ...

      env:
        AWS_REGION: ${{ fromJSON(inputs.variables).AWS_DEFAULT_REGION }}
        ...
```

klutchell:
My concern is how many repositories have custom actions that will need updating, and no one will know why they broke so they will ping us until they are all fixed. Definitely a major change, but one where it's on us to find and fix every repo.

ab77:
> My concern is how many repositories have custom actions that will need updating, and no one will know why they broke so they will ping us until they are all fixed. Definitely a major change, but one where it's on us to find and fix every repo.

I am fine with that.

klutchell:
Your commit should be changed to `major`.

These changes look fine to me, maybe open some draft PRs on other repos that will require compatibility updates.

At the very least run a search to get an idea of how many repositories we are talking about.

ab77:
> Your commit should be changed to `major`.
> 
> These changes look fine to me, maybe open some draft PRs on other repos that will require compatibility updates.
> 
> At the very least run a search to get an idea of how many repositories we are talking about.

Good news @klutchell : https://github.com/balena-io/balena-api/actions/runs/4297507190/jobs/7490593906#step:6:2

So looks like it's a warning instead of a hard fail.

klutchell:
> So looks like it's a warning instead of a hard fail. Do we still want this to be a "major"?

That is most excellent news! I guess it can be `minor` in that case since I bumped to v2 for self-hosted support already.

ab77:
lgtm

PR #546, opened 2023-03-10
Make sure protect branch always runs and cannot be skipped

klutchell:
A skipped or cancelled result will not prevent a PR from being merged.

Change-type: patch


klutchell:
I self-certify!

PR #551, opened 2023-03-14
Re-enable self-tests on armv7 and x86 self-hosted runners

klutchell:


klutchell:
> @klutchell the invariant tag is called `X64`, not `amd64`.

We are specifically looking for `platform:linux/amd64` which is added at build time. We don't need to use the GitHub automatic labels.

ab77:
@klutchell found it: https://github.com/organizations/balena-io/settings/actions/runner-groups/3

image

ab77:
@klutchell shod be good to go [now](https://github.com/product-os/flowzone/actions/runs/4419442080/jobs/7751496810#step:1:2)

The agents were failing self updates due to home directory permissions.

klutchell:
I self-certify!

PR #556, opened 2023-03-16
Remove auto-merge enable step

klutchell:
Also fix parsing of comma-separated inputs.

Change-type: patch


klutchell:
I self-certify!

PR #558, opened 2023-03-20
Remove manual docker cleanup steps

klutchell:
Runners should be ephemeral and these steps just add to build time.

Change-type: patch


klutchell:
@ab77 The ephemeral flags of the runners appear to be working, and we no longer need this step.

I verified on arm02 that the jammy runner restarted after this check run, as intended.

ab77 (approved):

PR #559, opened 2023-03-21
BREAKING: Change default docker tag format to {tag}-{variant}

klutchell:
This is a breaking change for any repositories expecting
image tags in the format `{variant}-{tag}`.

The new default aligns with [docker versioning](https://docs.renovatebot.com/modules/versioning/#docker-versioning).

The old behaviour can be restored with `docker_invert_tags: true`

Change-type: major

open-balena-base PR: https://github.com/balena-io-modules/open-balena-base/pull/271

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -108,6 +108,11 @@ on:
         type: string
         required: false
         default: default
+      docker_invert_tags:
+        description: Invert the tags for the Docker images (e.g. `{tag}-{variant}` becomes `{variant}-{tag}`)
+        type: boolean
+        required: false
+        default: false
       balena_environment:
         description: balenaCloud environment
         type: string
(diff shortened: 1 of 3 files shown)


ab77:
@klutchell is this something to do with handling our remote builders image names (focal-v1.2.3)?

klutchell:
> @klutchell is this something to do with handling our remote builders image names (focal-v1.2.3)?

Yup, this is that change you initially asked for to follow the Docker versioning syntax from renovate. I couldn't find a nice way to publish the new format without breaking existing projects, so this one can be turned on/off.

klutchell:
I'm also working on a new [PR](https://github.com/product-os/self-hosted-runners/pull/20) to publish `1.2.3-focal-arm32v7` and `1.2.3-jammy-arm32v8` etc. as it turns out to be very difficult to pull the same image name for different platforms in one compose file.

klutchell:
I self-certify!

PR #562, opened 2023-03-22
Run protect branch early in the workflow

klutchell:
Since we don't rely on Protect Branch as a required check anymore it actually saves time to run it early in the workflow, before jobs like All Tests and All Jobs have queued.

Change-type: minor


klutchell:
I self-certify!

PR #563, opened 2023-03-22
Reject duplicate workflow calls

klutchell:
This will force users to update their workflows
to avoid duplicate runs that could break branch protection.

Change-type: major


klutchell:
I self-certify!

PR #565, opened 2023-03-23
Allow publishing of intermediate docker images

klutchell:
Disabled by default because it's normally not necessary
and creates a lot of extra tags for multiarch images.

Change-type: minor


klutchell:
I self-certify!

PR #571, opened 2023-03-26
Reject internal pull requests using pull_request_target

klutchell:
This will prevent duplicate runs when both triggers are enabled.

Change-type: patch


klutchell:
I self-certify!

Issue #576, opened 2023-03-28
docker label "org.opencontainers.image.ref.name" is just the build context dir


klutchell:
We should either disable generation of the label in `test_meta` or set it to something more appropriate.

See: https://github.com/docker/metadata-action
See: https://github.com/product-os/self-hosted-runners/pkgs/container/self-hosted-runners/80735027?tag=1.2.8-arm64v8
```json
"labels": {
    "org.opencontainers.image.url": "https://github.com/product-os/self-hosted-runners",
    "org.opencontainers.image.licenses": "Apache-2.0",
    "org.opencontainers.image.title": "self-hosted-runners",
    "org.opencontainers.image.revision": "f828e8a8f0b96cfef89b3659eea55db60e47849f",
    "org.opencontainers.image.created": "2023-03-27T18:29:07.834Z",
    "org.opencontainers.image.version": "1.2.8",
    "org.opencontainers.image.description": "GitHub self-hosted runner images for common system architectures.",
    "org.opencontainers.image.ref.name": "ubuntu",
    "org.opencontainers.image.source": "https://github.com/product-os/self-hosted-runners"
  }
```
The bake target might be a good option.

PR #577, opened 2023-03-29
Use the bake target for the ref.name docker image label

klutchell:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/576


klutchell:
I self-certify!

PR #579, opened 2023-03-30
Add app token generation to custom action examples

klutchell:
Change-type: patch


ab77 (approved):

klutchell:
@ab77 the other option here is have Flowzone always generate an ephemeral token and pass it to the custom actions as a new input. The downside being the custom action may not need that token, and might want to perform some long running tasks before requesting a token.

It seems better to have the custom actions generate their own token if they need one, following the examples.

PR #585, opened 2023-04-03
Disable auto-merge by default

klutchell:
This seems to be blocking PR close events due to some kind of race condition.

Change-type: patch


klutchell:
I self-certify!

PR #586, opened 2023-04-03
Do not use automatic GHA token to enable auto-merge

klutchell:
The automatic GHA token intentionally does not trigger additional actions events, so using it to enable auto-merge causes PR merge events not to trigger any actions.

Change-type: patch


klutchell:
I self-certify!

Issue #582, opened 2023-04-03
Skip npm pack for projects where private=true in package.lock


klutchell:

klutchell:
Resolved by https://github.com/product-os/flowzone/pull/588

PR #588, opened 2023-04-04
Skip npm pack for private packages

klutchell:
Change-type: patch
Resolves: https://github.com/product-os/flowzone/issues/582


klutchell:
I self-certify!

PR #590, opened 2023-04-04
Combine tagged source and versioned source jobs

klutchell:
These jobs were performing nrealy the same functions and combining them reduces the paths required for testing.

Change-type: patch


klutchell:
I self-certify!

PR #593, opened 2023-04-04
Remove wait for required checks

klutchell:
This job was causing dependency loops for any required
job that dependend on Flowzone.

For now we will see if the GH auto-merge toggle behaves as designed.

Change-type: patch


klutchell:
I self-certify!

PR #594, opened 2023-04-04
Toggle auto-merge with a reduced scope ephemeral token

klutchell:
GitHub support suggested this should prevent auto-merging without required checks.

Change-type: patch


klutchell:
I self-certify!

PR #596, opened 2023-04-05
Use automatic GHA token for steps with user code

klutchell:
This is a cleanup of the logic to determine which
GH token to provide in different situations.

In all cases we should use the GH app token, or the FLOWZONE_TOKEN, except for user code.

Change-type: patch


klutchell:
I self-certify!

PR #602, opened 2023-04-10
Allows each repo to customize extend-ignore rules on python

otaviojacobi:
Change-type: patch

Similar to ts-linter in Python we can use flake8 to lint our code. The problem is that we currently pass the extend-ignore by parameter, which blocks each repo of ignoring its own rules (ideally we want to use all flake8 rules but in reality each repo may sometimes want to ignore particular rules, to do so, one can simply add a flake8 rule to the poetry file or a .flake8 file which will be used)

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -2398,7 +2398,7 @@ jobs:
           fi
       - name: Lint with flake8
         run: |
-          poetry run flake8 --max-line-length=120 --benchmark --extend-ignore=E203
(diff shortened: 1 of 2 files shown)


otaviojacobi:
LGTM

gantonayde:
@otaviojacobi Wouldn't that make all Python projects without a .flake8 file fail the test?

otaviojacobi:
@gantonayde only if they don't respect the rule :)

PR #603, opened 2023-04-11
Publishes python(poetry) packages to PyPI

otaviojacobi:
Change-type: minor

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,8 @@ Python tests will be run if a `pyproject.toml` file is found in the root of the 
 
 Multiple versions (>=3.7, <3.11) of Python will be tested as long as they meet the range in the `pyproject.toml` file.
 
+If your `pyproject.toml` file contains a `packages` property under `[tool.poetry]` tag, this package will also be deployed to [PyPI registry](https://pypi.org/) on master and [PyPI test registry](https://test.pypi.org/) on PRs.
+
 ### Rust
 
 If a `Cargo.toml` file is found in the root of the repository, Flowzone will run tests on the code formatting (using `cargo fmt` and `cargo clippy`) and then run tests for a set of target architectures given in [`cargo_targets`](#cargo_targets). In order to disable Rust testing, set the value of that variable to `""`.
@@ -399,6 +401,18 @@ Required for [Docker](#docker) projects.
 [API key](https://www.balena.io/docs/learn/manage/account/#api-keys) for pushing releases to balena applications.
 
 Required for [balena](#balena) projects.
+
+#### `PYPI_TOKEN`
+
+[API key](https://pypi.org/help/#apitoken) for pushing releases to PyPI applications.
+
+Used for [python](#python) projects.
+
+#### `PYPI_TEST_TOKEN`
+
+[API key](https://pypi.org/help/#apitoken) for pushing releases to test PyPI applications.
+
+Used for [python](#python) PR artifacts.
 
 #### `CARGO_REGISTRY_TOKEN`
 
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -54,6 +54,22 @@
       echo "semver=$(npx -q -y -- semver -c -l "${tag}")" >> $GITHUB_OUTPUT
       echo "describe=$(git describe --tags --always --dirty | cat)" >> $GITHUB_OUTPUT
 
+
+  - &generatePythonMetadata
+    name: Generate Python metadata
+    id: python_meta
+    run: |
+      package=$(poetry version --no-ansi | awk '{print $1}')
+      version=$(poetry version --no-ansi | awk '{print $2}')
+      commit_sha="$(echo ${{ github.event.pull_request.head.sha }} | tr "a-z" "A-Z")"
+      decimal_sha=$(echo "ibase=16; $commit_sha" | bc)
+      version_tag="${version}-dev${decimal_sha}"
+
+      echo "package=${package}" >> $GITHUB_OUTPUT
+      echo "version=${version}" >> $GITHUB_OUTPUT
+      echo "version_tag=${version_tag}" >> $GITHUB_OUTPUT
+
+
   - &deployToBalenaAction
     uses: balena-io/deploy-to-balena-action@e7041b7ca6dd85f47a3eb5ce141178bf77d85920 # v0.27.0
     id: balena_deploy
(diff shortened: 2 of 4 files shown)


flowzone.yml
```diff
@@ -2335,6 +2335,81 @@ jobs:
...
+      - name: Run poetry install
+        run: |
+          poetry install
+
+      - name: Publish release
+        env:
+          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+        run: |
+          poetry config pypi-token.pypi $PYPI_TOKEN
+          poetry publish --build
```
klutchell, line 2411: Generally for finalize steps we do not want to build anything new, and instead try to re-tag or re-upload something built in a previous publish or test step.

For example, docker finalize and balena finalize both just re-tag existing draft artifacts.

NPM finalize cannot re-tag so it downloads the pack from the previous publish run and re-uploads that.

Can pypi.org packages be copied from test.pypi.org to finalize? Or is there some other workflow that would avoid building new artifacts? Maybe following the npm publish & finalize example?

otaviojacobi: hey @klutchell this is still WIP and I am still learning the working of PyPI in more details. After a small struggle this morning using test PyPI I managed to release a pre-release into test PyPI which means I can also do a pre-release on the actual PyPI - so we would probably be able to ditch Test PyPI all together and use only the PyPI (non-test) registry, which would allow for what you described.

FWIW the struggle here is because PyPI does not actually use full semver specification but rather PEP440 which is slightly different: https://peps.python.org/pep-0440/

More over, I can only do pre-releases using the commit hash "small version" such as https://test.pypi.org/project/balena-sdk-python/12.6.0b445977/

klutchell: Sounds like you are on the right track, I'll leave you to it!

gantonayde:
@otaviojacobi You might want to check this out https://github.com/nektos/act . Carlo mentioned he used it when testing the github actions for the support scheduler bot as it allows you to run an action locally. During development, it becomes very tedious very quickly when you need to re-push for every little change (don't ask me how I know :D)

klutchell:
> @otaviojacobi You might want to check this out https://github.com/nektos/act . Carlo mentioned he used it when testing the github actions for the support scheduler bot as it allows you to run an action locally. During development, it becomes very tedious very quickly when you need to re-push for every little change (don't ask me how I know :D)

See this comment regarding the work Miguel did to get Flowzone working with ACT.
https://github.com/product-os/flowzone/issues/63#issuecomment-1246199288

README.md
```diff
@@ -400,6 +402,16 @@ Required for [Docker](#docker) projects.
 
 Required for [balena](#balena) projects.
 
+#### `PYPI_TOKEN`
+[API key](https://pypi.org/help/#apitoken) for pushing releases to PyPI applications.
+
+Used for [python](#python) projects.
+
+#### `PYPI_TEST_TOKEN`
+[API key](https://pypi.org/help/#apitoken) for pushing releases to test PyPI applications.
```
klutchell, line 411: missing newline for Markdown syntax (try a markdown lint vscode extension)

README.md
```diff
@@ -400,6 +402,16 @@ Required for [Docker](#docker) projects.
 
 Required for [balena](#balena) projects.
 
+#### `PYPI_TOKEN`
```
klutchell, line 405: missing newline for Markdown syntax (try a markdown lint vscode extension)

flowzone.yml
```diff
@@ -54,6 +54,23 @@
...
 
+
+  - &generatePythonMetadata
+    name: Generate Python metadata
+    id: python_meta
+    run: |
+      package=$(poetry version --no-ansi | awk '{print $1}')
+      version=$(poetry version --no-ansi | awk '{print $2}')
+      commit_sha=${{ github.event.pull_request.head.sha }}
+      commit_sha=$(echo $commit_sha | tr "a-z" "A-Z")
```
klutchell, line 65: These two lines can be combined, and still be readable I think.
```suggestion
      commit_sha="$(echo ${{ github.event.pull_request.head.sha }} | tr "a-z" "A-Z")"
```

flowzone.yml
```diff
@@ -1305,8 +1323,18 @@ jobs:
               echo "Poetry not used"
               echo "enabled=false" >> $GITHUB_OUTPUT
             fi
+
+            has_package=$(awk -F "=" '/^packages/ {print $2}' pyproject.toml)
```
klutchell, line 1327: Is it possible for this property to have multiple values? How is that handled?

README.md
otaviojacobi: Done

README.md
otaviojacobi: done

flowzone.yml
otaviojacobi: done

flowzone.yml
```diff
@@ -1285,6 +1302,7 @@ jobs:
     outputs:
       python_poetry: ${{ steps.python_poetry.outputs.enabled }}
       python_versions: ${{ steps.python_versions.outputs.json }}
+      pypi_deployment: ${{ steps.python_poetry.outputs.pypi_deployment }}
```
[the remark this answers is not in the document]
otaviojacobi: Agreed,  pypi publish is far more readable

flowzone.yml
otaviojacobi: @klutchell  It is - but this just means "all these subfolders/modules are part of the final package to be deployed" which means, poetry will handle bundling all those into a wheel/tgz when we call `poetry publish --build` or `poetry build`. Now the problem I see is that we don't want to deploy every single python/poetry repo we have. My workaround this problem was to use packages (which is usually only used if you want to deploy something and I did not found something like "private" in NPM). I found a single repo of our internal python stuff we don't want to publish and has a "packages", so I need to talk to Genadi regarding that before merging this PR. 

If you have other suggestions of how we could figure it out if the package should/should not be deployed I am open to it too :)

klutchell: Yeah, I don't see any other fields in pyproject.toml that would make sense. I guess the only other option is to add a new boolean to Flowzone. Do we want to reach out for more ideas/opinions first?

klutchell:
Is this ready for final review?

otaviojacobi:
> Is this ready for final review?

@klutchell I would like to check that this works on the SDK repo before the final touches, but I need the environments to be setup there. I pinged you in zullip with the thread

klutchell (approved):

PR #609, opened 2023-04-14
Adds patch bump to commit footer on e2e

otaviojacobi:
Change-type: patch

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -183,7 +183,7 @@ jobs:
           yq '.jobs.flowzone.uses |= . line_comment="${{ github.run_id }}-${{ github.run_attempt }}"' -i "${WORKFLOW_FILE}"
 
           git add "${WORKFLOW_FILE}"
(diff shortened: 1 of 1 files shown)


klutchell:
@otaviojacobi you can test this by running the e2e workflow on your python-sdk repo, but in the workflow command line add `--ref otaviojacobi/adds-patch-change-to-footer` so the workflow itself uses this branch of `e2e.yml`.

otaviojacobi:
> @otaviojacobi you can test this by running the e2e workflow on your python-sdk repo, but in the workflow command line add `--ref otaviojacobi/adds-patch-change-to-footer` so the workflow itself uses this branch of `e2e.yml`.

@klutchell I will do that next week, for the time being I can test because I did some small fix on the sdk repo versionist file to support patch: etc

otaviojacobi:
Seems to work, please see: https://github.com/balena-io/balena-sdk-python/pull/305/commits

klutchell:
lgtm

Issue #548, opened 2023-03-13
Disable provenance attestation in docker bake action


klutchell:
See:
- https://github.com/docker/bake-action/pull/104
- https://www.docker.com/blog/highlights-buildkit-v0-11-release/
- https://docs.docker.com/build/attestations/attestation-storage/#attestation-manifest
- https://github.com/docker/build-push-action/pull/746
- https://github.com/docker/buildx/issues/1509
- https://github.com/docker/build-push-action/issues/764

klutchell:
Resolved by https://github.com/product-os/flowzone/pull/612

PR #612, opened 2023-04-18
Disable SLSA Provenance attestation for now

klutchell:
Resolves: https://github.com/product-os/flowzone/issues/548
See: https://github.com/docker/bake-action/releases/tag/v3.0.0
See: https://slsa.dev/provenance/v0.2

We can re-enable in the future when support is more common.

Change-type: patch


klutchell:
I self-certify!

PR #623, opened 2023-04-25
Only pack and upload for the latest node version

klutchell:
Change-type: minor

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -54,7 +54,6 @@
       echo "semver=$(npx -q -y -- semver -c -l "${tag}")" >> $GITHUB_OUTPUT
       echo "describe=$(git describe --tags --always --dirty | cat)" >> $GITHUB_OUTPUT
 
-
   - &generatePythonMetadata
     name: Generate Python metadata
     id: python_meta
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -499,6 +497,15 @@
...
       echo "block_creations__enabled=$(jq -cr '.block_creations.enabled // false' <<< "${result}")" >> $GITHUB_OUTPUT
 
+  - &sortNodeVersions
+    name: Sort node version
+    id: node_versions
+    env:
+      VERSIONS: ${{ needs.is_npm.outputs.node_versions }}
+    run: |
+      echo "min=$(echo "${VERSIONS}" | jq -r '.[]' | sort -V | head -n1)" >> $GITHUB_OUTPUT
+      echo "max=$(echo "${VERSIONS}" | jq -r '.[]' | sort -V -r | head -n1)" >> $GITHUB_OUTPUT
```
Page-, line 507: In general I prefer long flags for committed scripts so that it's easier for people who are later reading it and may not be as familiar with the short flags:
```suggestion
      echo "min=$(echo "${VERSIONS}" | jq -r '.[]' | sort --version-sort | head -n1)" >> $GITHUB_OUTPUT
      echo "max=$(echo "${VERSIONS}" | jq -r '.[]' | sort --version-sort --reverse | head -n1)" >> $GITHUB_OUTPUT
```

klutchell:
I self-certify!

PR #632, opened 2023-05-08
Write `script` output to /tmp

pipex:
The `script` command used for pseudo-tty during testing writes to `typescript` in the current directory by default. This may cause issues with some projects that do not expect new files to be generated. This changes the location of that file to the `/tmp` directory to avoid any issues.

Change-type: patch


pipex:
I self-certify!

PR #630, opened 2023-05-05
Enable pseudo-tty allocation on test steps

pipex:
This wraps the shell in the steps that run `npm run test`
and `poetry run pytest` in a `script -q -e -c` to allow tests that
require pseudo-tty (e.g. for applications that work with terminal colors)
to be able to operate as they would in a local dev environment.

Change-type: minor

PR #633, opened 2023-05-09
Make pseudo-tty allocation optional.

pipex:
Pseudo-tty allocation uses the `script` command for setting up the environment, however this really only works in linux runners, meaning users trying to setup runs on other environments will see their tests fail for unrelated reasons. This makes the configuration optional behind a `pseudo_terminal` input.

Change-type: minor
Relates-to: #630


klutchell:
I think it would be better to expose a `default_shell` input and use the same default we have today.

README.md
```diff
@@ -598,6 +599,14 @@ Type: _string_
 
 Default: `'stable'`
 
+#### `pseudo_terminal`
+
+Set to true to enable terminal emulation for test steps (under `npm_test` and `python_tests`). This wraps the test call with the `script` command so test commands that rely on a terminal feature (e.g. terminal colors) can work as they would in a development environment. This will only works with ubuntu runners.
```
klutchell, line 604:
```suggestion
Set to true to enable terminal emulation for test steps (under `npm_test` and `python_tests`). This wraps the test call with the `script` command so test commands that rely on a terminal feature (e.g. terminal colors) can work as they would in a development environment. This only works with ubuntu runners.
```

pipex:
I self-certify!

PR #636, opened 2023-05-11
Revert "Convert artifact compression to zstd"

fisehara:
Reverts product-os/flowzone#635

Jenkins Runners don't have zstd installed and aarch64 runners are stuck on old releases, so that they don't pick up the changes made on the jenkins-runners latest release


klutchell:
The self-hosted runners have been updated, and all hosted runners are supported except for macos-11.

PR #639, opened 2023-05-19
Use elevated tokens for deploy to balena registry secrets

klutchell:
The deploy slug is trusted code in the flowzone.yml so we can trust that private images will only be deployed to trusted fleets/apps.

Change-type: patch


klutchell:
Converting to draft as currently app installation tokens do not have access to organization packages.

ab77 (approved):

PR #640, opened 2023-05-23
inject AWS session credentials (OIDC)

ab77:
* optionally add AWS auth to .env compose file to enable aws-cli operations in docker compose workflows
* requires IAM configuration (e.g. https://github.com/product-os/environment-production/pull/553)

Tested: https://github.com/product-os/environment-production/actions/runs/5060183471/jobs/9083089191

change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -568,6 +568,16 @@ on:
         required: false
 
     inputs:
+      aws_region:
+        description: "AWS region with GitHub OIDC provider IAM configuration"
+        type: string
+        required: false
+        default: "${{ vars.AWS_REGION || '' }}"
+      aws_iam_role:
+        description: "AWS IAM role ARN to assume with GitHub OIDC provider"
+        type: string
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -1919,6 +1929,10 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.is_docker.outputs.docker_test_matrix) }}
 
+    # https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings
+    permissions:
+     id-token: write
```
klutchell, line 1934: > You may need to specify additional permissions here, depending on your workflow's requirements.

I think this removes almost all permissions from the GITHUB_TOKEN secret we are passing into docker bake as a build secret. Since we don't know how that build arg is used we should try to keep the permissions the same and just add what we need.

PR #655, opened 2023-06-13
Create versioned commit objects via GitHub API

klutchell:
These commit objects will be automatically signed and verified
if using GitHub app tokens for authentication.

This also removes the need for versioned source artifacts
as each job can clone the versioned commit directly.

Change-type: major


klutchell:
e2e results: https://github.com/product-os/flowzone/actions/runs/5270358695

klutchell:
I self-certify!

PR #664, opened 2023-06-15
Do not use github.sha for pull request target events

klutchell:
For PRT this sha is from master, not the PR branch.

Change-type: patch


klutchell:
I self-certify!

PR #665, opened 2023-06-15
Make protect branch depend on versioned source

klutchell:
If only to make sure we don't apply new branch protections when the commits don't pass the scrutiny of versionist and rebase requirements.

Change-type: patch


klutchell:
I self-certify!

PR #673, opened 2023-06-20
Terraform project type

ab77:
Add `terraform` project type and configures with TF state stored in Kubernetes cluster(s) (e.g.):

```
      terraform_projects: >
        {
          "projects": [
            {
              "main_tf": "terraform/github"
            },
            {
              "main_tf": "terraform/authentication"
            },
            ...
          ]
        }
```

Test: https://github.com/product-os/environment-staging/actions/runs/5326084442/jobs/9648989905

* also includes reliability settings for AWS/CFN and TF jobs

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -529,6 +529,50 @@
     name: Get caller identity (AWS/whoami)
     run: aws sts get-caller-identity
 
+  - &updateKubeconfig
+    name: Update kubeconfig
+    run: |
+      aws eks update-kubeconfig --name $(echo "${KUBE_CTX}" | awk -F'/' '{print $2}')
+
+  - &convenienceFunctions
+    name: Convenience functions
+    id: functions
+    run: |
+      EOF="$(openssl rand -hex 16)"
+
+      # https://sre.google/sre-book/addressing-cascading-failures/
+      with_backoff="$(mktemp)"
+      cat << $EOF > "${with_backoff}"
+      function with_backoff() {
+          local max_attempts=\${ATTEMPTS-3}
+          local timeout=\${TIMEOUT-2}
+          local attempt=0
+          local exitCode=0
+          set +e
+          while [[ \$attempt < \$max_attempts ]]; do
+              "\$@"
+              exitCode=\$?
+              [[ \$exitCode == 0 ]] && break
+              echo "Failure! Retrying in \$timeout.." 1>&2
+              sleep "\$timeout"
+              attempt=\$(( attempt + 1 ))
+              timeout=\$(( timeout * 2 ))
+          done
+          [[ \$exitCode != 0 ]] && echo "You've failed me for the last time! (\$*)" 1>&2
+          set -e
+          return \$exitCode
+      }
+      $EOF
+      echo "with_backoff=${with_backoff}" >> $GITHUB_OUTPUT
+
+  - &randomDelay
+    name: Random delay
+    run: |
+      DELAY=${DELAY-5}
+      random=$(((RANDOM % DELAY) + 1))
+      echo "::notice::sleeping for ${random}s"
+      sleep "${random}s"
+
 name: Flowzone
 
 on:
@@ -596,7 +640,50 @@ on:
         required: false
         default: "${{ vars.AWS_IAM_ROLE || '' }}"
       cloudformation_templates:
-        description: "AWS CloudFormation templates to deploy"
+        description: |
+          AWS CloudFormation templates to deploy (e.g.)
+
+          ```
+          {
+            "stacks": [
+              {
+                "name": "foo",
+                "template": "aws/bar.yaml",
+                "tags": [
+                  "Name=foo",
+                  "Environment=${FOO}"
+                ],
+                "capabilities": [
+                  "CAPABILITY_IAM",
+                  "CAPABILITY_NAMED_IAM"
+                ]
+              },
+              ...
+            ]
+          }
+          ```
+
+          * assumes `aws/bar.yaml` exists.
+          * `${ENVVARS}` injected at runtime from `vars` and `secrets` contexts
+        type: string
+        required: false
+        default: ""
+      terraform_projects:
+        description: |
+          Terraform projects to deploy (e.g.)
+
+          ```
+          {
+            "projects": [
+              {
+                "main_tf": "terraform/foo"
+              }
+            ]
+          }
+          ```
+
+          * assumes `terraform/foo/main.tf` exists
+          * GitHub `vars` and `secrets` contexts injected into TF plan/apply steps
         type: string
         required: false
         default: ""
@@ -1744,8 +1831,7 @@ jobs:
     if: |
       !failure() && !cancelled() &&
       needs.versioned_source.result == 'success' &&
-      inputs.cloudformation_templates != '' &&
-      github.event.pull_request.head.repo.full_name == github.repository
+      inputs.cloudformation_templates != ''
 
     <<: *customWorkingDirectory
 
@@ -1759,10 +1845,16 @@ jobs:
       - name: Validate CloudFormation input
         id: validate_json
         run: |
-          if echo '${{ inputs.cloudformation_templates }}' | jq -r
+          if echo '${{ inputs.cloudformation_templates }}' | jq -r .stacks
           then
+            if [[ '${{ github.event.pull_request.head.repo.full_name }}' != '${{ github.repository }}' ]]; then
+              echo '::warning:: CloudFormation stacks are skipped for external contributions.'
+              echo "enabled=false" >> $GITHUB_OUTPUT
+              exit 0
+            fi
             echo "enabled=true" >> $GITHUB_OUTPUT
           else
+            echo '::warning::invalid JSON or no CloudFormation stacks?'
             echo "enabled=false" >> $GITHUB_OUTPUT
           fi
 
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -600,6 +605,11 @@ on:
         type: string
         required: false
         default: ""
+      terraform_projects:
+        description: "Terraform projects to deploy"
+        type: string
+        required: false
+        default: ""
```
klutchell, line 612: If this needs to be in a certain format, like JSON, or comma separated, we should specify that in the description.

Be as verbose as possible in the description since it is applied to the readme, and multiple lines are supported.

flowzone.yml
```diff
@@ -1773,6 +1783,44 @@ jobs:
...
+      github.event.pull_request.head.repo.full_name == github.repository
+
+    <<: *customWorkingDirectory
+
+    outputs:
+      terraform: ${{ steps.validate_json.outputs.enabled }}
+      projects: ${{ steps.terraform_projects.outputs.deploy }}
+
+    steps:
+      - *checkoutVersionedSha
```
klutchell, line 1805:
```suggestion
```
Seems like we don't actually use the source here.

flowzone.yml
```diff
@@ -3312,6 +3360,86 @@ jobs:
...
+  # Terraform
+  ###################################################
+
+  terraform_test:
+    name: Test Terraform
+    runs-on: ${{ fromJSON(inputs.runs_on) }}
+    strategy:
+      fail-fast: true
+      matrix:
+        project: ${{ fromJSON(needs.is_terraform.outputs.projects) }}
```
klutchell, line 3373:
```suggestion
        project: ${{ fromJSON(needs.is_terraform.outputs.deploy) }}
```
You've named your output `deploy` on line 1822

flowzone.yml
```diff
@@ -3312,6 +3360,86 @@ jobs:
...
+    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
+    needs:
+      - is_pr_open
+      - is_terraform
+      - versioned_source
+    if: |
+      !failure() && !cancelled() &&
+      needs.is_pr_open.result == 'success' &&
+      needs.is_terraform.outputs.terraform == 'true'
```
klutchell, line 3382:
```suggestion
      needs.is_terraform.outputs.enabled == 'true'
```

flowzone.yml
```diff
@@ -3312,6 +3360,86 @@ jobs:
...
+        project: ${{ fromJSON(needs.is_terraform.outputs.projects) }}
+    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
+    needs:
+      - is_pr_merged
+      - is_pushed_tag
+      - is_terraform
+    if: |
+      !failure() && !cancelled() &&
+      needs.is_pr_merged.result != needs.is_pushed_tag.result &&
+      needs.is_terraform.outputs.terraform == 'true'
```
klutchell, line 3425:
```suggestion
      needs.is_terraform.outputs.enabled == 'true'
```

flowzone.yml
```diff
@@ -3312,6 +3360,86 @@ jobs:
...
+
+    defaults:
+      run:
+        working-directory: ${{ matrix.project }}
+
+    env:
+      KUBE_CTX: ${{ vars.KUBECTL_CONTEXT }}
+      KUBE_NAMESPACE: ${{ vars.KUBE_NAMESPACE }}
+
+    steps:
```
klutchell, line 3435: Is this step missing `- *checkoutVersionedSha` ?

flowzone.yml
ab77:
```
  is_terraform:
    ...
    outputs:
      terraform: ${{ steps.validate_json.outputs.enabled }}
```

So this should be correct..

flowzone.yml
ab77: ditto

flowzone.yml
```diff
@@ -1773,6 +1820,43 @@ jobs:
...
+    name: Is Terraform
+    runs-on: ${{ fromJSON(inputs.runs_on) }}
+    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
+    needs:
+      - versioned_source
+    if: |
+      !failure() && !cancelled() &&
+      needs.versioned_source.result == 'success' &&
+      inputs.terraform_projects != '' &&
+      github.event.pull_request.head.repo.full_name == github.repository
```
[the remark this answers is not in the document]
ab77: We are just testing for valid JSON here, that's about it. Can print something like `echo '::warning::invalid JSON?'`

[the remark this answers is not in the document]
klutchell: I mean move the check `github.event.pull_request.head.repo.full_name == github.repository` into the `Validate Terraform input` step so we can print a warning and not return `enabled=true`.

flowzone.yml
```diff
@@ -1773,6 +1822,44 @@ jobs:
...
+      terraform: ${{ steps.validate_json.outputs.enabled }}
+      projects: ${{ steps.terraform_projects.outputs.deploy }}
+
+    steps:
+
+      - name: Validate Terraform input
+        id: validate_json
+        run: |
+          if echo '${{ inputs.terraform_projects }}' | jq -r .projects
+          then
```
klutchell, line 1849:
```suggestion
          then
            if [ "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ; then
              echo "warning:: Terraform projects are skipped for external contributions."
              exit 0
            fi
```

ab77:
@cmfcruz Can you please check this, when you get chance?

flowzone.yml
```diff
@@ -3555,6 +3850,7 @@ jobs:
       - custom_publish
       - custom_always
       - cloudformation_finalize
+      - terraform_finalize
```
klutchell, line 3853: These finalize jobs don't need to be here, since the requirement for All Jobs really only matters for PR branch rules.

I don't _think_ it will break, but it is certainly not required.

However `cloudformation_test` and `terraform_test` should be here. Maybe add a comment explaining the above so others don't make the same mistake.

flowzone.yml
```diff
@@ -3145,17 +3267,26 @@ jobs:
...
       - name: Create templates bucket
         id: make_bucket
         run: |
-          bucket="$(aws s3api list-buckets | jq -r '.Buckets[] | select(.Name=="${{ vars.AWS_S3_BUCKET }}").Name' || true)"
+          source '${{ steps.functions.outputs.with_backoff }}'
+
+          bucket="$(with_backoff aws s3api list-buckets | jq -r '.Buckets[] | select(.Name=="${{ vars.AWS_S3_BUCKET }}").Name' || true)"
           if [[ -z "$bucket" ]]; then
-              result="$(aws s3 mb "s3://cfn-$(uuidgen)-${{ inputs.aws_region }}" --region '${{ inputs.aws_region }}')"
+              result="$(with_backoff aws s3 mb "s3://cfn-$(uuidgen)-${{ inputs.aws_region }}" --region '${{ inputs.aws_region }}')"
```
klutchell, line 3289: What is so unreliable about these aws calls that we need a retry backoff all over the place?

I hated that part of the remote-builder tests, are we using these tools incorrectly that they always need multiple attempts?

flowzone.yml
```diff
@@ -1773,6 +1855,46 @@ jobs:
...
+
+    steps:
+      - name: Validate Terraform input
+        id: validate_json
+        run: |
+          if echo '${{ inputs.terraform_projects }}' | jq -r .projects
+          then
+            if [[ '${{ github.event.pull_request.head.repo.full_name }}' != '${{ github.repository }}' ]]; then
+              echo '::warning:: Terraform projects are skipped for external contributions.'
+              exit 0
```
klutchell, line 1883:
```suggestion
              echo "enabled=false" >> $GITHUB_OUTPUT
              exit 0
```
Either set it to false in both failed conditions, or never set it to false and only check for true conditions. Not a mix of both.

flowzone.yml
ab77: @klutchell AWS have rate limits, so if 10 requests arrive from the same IP within a short time, jobs fail.

flowzone.yml
ab77: @klutchell before I added `cloudformation_finalize`, my PRs would merge prematurely while those jobs were still running. So at least one of these should be there for each project type, or I've missed something..

flowzone.yml
ab77: I don't follow. The current logic would not set step outputs if the request is coming from a fork.

flowzone.yml
klutchell: the `cloudformation_test` job should be there, same for terraform, because finalize only happens after the PR is merged, why would we want to block the merge waiting for finalize jobs?

flowzone.yml
klutchell: we should add a comment explaining that or someone will delete the retries, probably me when I forget why we added them

flowzone.yml
klutchell: All Jobs only exists so we can make it a required check for merges. It doesn't make any sense for post-merge jobs to be there, but all the test and deploy jobs should be there.

flowzone.yml
klutchell: I know, but there is another condition with `echo '::warning::invalid JSON or no CloudFormation stacks?'` where we are setting it explicitly to false.

All I'm saying is be consistent. Either set it to false, or don't set it. But be consistent.

flowzone.yml
ab77: Good read: https://sre.google/sre-book/addressing-cascading-failures/

klutchell: Yeah, I've seen that. We should try this action sometime:
https://github.com/marketplace/actions/retry-step

It doesn't have automatic random backoffs, but we could add our own randomization with `on_retry_command`.

klutchell: Either way, a comment would be nice to explain why this is needed here.

ab77:
lgtm

PR #681, opened 2023-06-28
Create local refs for orphaned commit and tag objects

klutchell:
For open pull requests, the versioned commit and tag
objects are created but are intentionally not referenced
anywhere in the git tree.

As such, command line tools like git describe do not return
the current versioned tag which is sometimes used in build scripts.

So before build, test, and publish jobs we can create the references
to the orphaned objects locally. These references should already
exist on merge/finalize.

Change-type: minor


klutchell:
I self-certify!

PR #684, opened 2023-06-28
Disable GPT reviews for Flowzone self-tests for now

klutchell:
There are some outstanding issues that should be resolved before enabling this feature anywhere.

Change-type: patch


klutchell:
I self-certify!

PR #688, opened 2023-07-11
Get Github app token before cloning versioned source

klutchell:
Otherwise we blindly fall back to FLOWZONE_TOKEN with no warning

Change-type: patch


klutchell:
I self-certify!

PR #693, opened 2023-07-18
Use gh api command to create blobs

klutchell:
This command is more likely to catch errors that are missed by curl.

Change-type: patch


klutchell:
I self-certify!

PR #694, opened 2023-07-18
Fix concurrency for external contributions and non-syncronize events

klutchell:
Use the PR number for uniqueness when determining concurrency, and only cancel in progress jobs if the PR is open.

Change-type: patch


klutchell:
I self-certify!

PR #696, opened 2023-07-31
Remove the open condition from external PR checks

klutchell:
In the current use cases it doesn't matter if the PR is open or not when performing the steps, and in some cases like configureAWSCredentials we actually expect the PR to be closed sometimes.

Change-type: patch


klutchell:
I self-certify!

PR #702, opened 2023-08-02
Use `--locked` flag when installing cross

pipex:
This is to prevent bugs from dependencies from breaking the build.

See: https://github.com/balena-io-modules/node-systemd/actions/runs/5740095381/job/15557192112
Caused by: https://github.com/bytecodealliance/rustix/issues/767

Change-type: patch


pipex:
I self-certify!

PR #704, opened 2023-08-04
Only provide automatic GITHUB_TOKEN to compose vars

klutchell:
The docker compose tests are one of the few places where we are providing elevated secrets to untrusted user code beyond the workflow itself.

In the case of a GitHub token appended to compose vars, we should use the automatic GHA token if possible and avoid leaking a privileged PAT or ephemeral app token, even when restricted to internal PRs.

Change-type: minor
See: https://github.com/product-os/self-hosted-runners/pull/75


klutchell:
I self-certify!

PR #706, opened 2023-08-08
Test all job types on self-hosted runners

klutchell:
This will catch any missing dependencies or configuration on our self-hosted runners required to completely
avoid GitHub hosted runners on private repos.

Change-type: patch


klutchell:
I self-certify!

Issue #703, opened 2023-08-04
Review supported Python versions


pranasziaukas:
As of today, Python 3.8-3.11 is actively supported, see https://devguide.python.org/versions

We may want to consider dropping references to Python 3.7 and including Python 3.11.

Relevant code:
https://github.com/product-os/flowzone/blob/b2d6f37f6c40e0b10c3611d4af5d0ad5724ed4bd/.github/workflows/flowzone.yml#L1015-L1074

PR #709, opened 2023-08-10
Implement generic Python version compatibility checks

pranasziaukas:
We duplicate quite a bit of code in order to support different Python versions. This PR proposes a way to only specify supported versions once (as a variable), and the `is_python` job takes care of everything else.

It relies on the fact that the `setup-python` action supports simultaneous installation of multiple versions (documentation [here](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-multiple-pythonpypy-version)), and then `poetry env use` can be invoked as part of a loop.

Also happens to address #703

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1483,7 +1483,13 @@ jobs:
           echo "build=${matrix}">> $GITHUB_OUTPUT
 
   is_python:
-    name: Is python
+    name: Determine compatible Python versions
+    env:
+      SUPPORTED_VERSIONS: |
+        3.8
+        3.9
+        3.10
+        3.11
     runs-on: ${{ fromJSON(inputs.runs_on) }}
     timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
     needs:
@@ -1500,7 +1506,7 @@ jobs:
       - *getGitHubAppToken
       - *checkoutVersionedSha
 
-      - name: Check for Python Poetry pyproject.toml
+      - name: Identify Poetry project
         id: python_poetry
         run: |
           if test -f "pyproject.toml"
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1483,7 +1483,13 @@ jobs:
...
 
   is_python:
-    name: Is python
+    name: Determine compatible Python versions
+    env:
+      SUPPORTED_VERSIONS: |
+        3.8
+        3.9
+        3.10
+        3.11
```
pranasziaukas, line 1492: The single source of truth.

flowzone.yml
```diff
@@ -1529,87 +1535,39 @@ jobs:
...
-            echo "Python 3.7 does not meet project requirements."
-          fi
-
-      - name: Set up Python 3.8
+      - name: Set up Python environment
         if: steps.python_poetry.outputs.enabled == 'true'
         uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4
         with:
-          python-version: "3.8"
+          python-version: ${{ env.SUPPORTED_VERSIONS }}
```
pranasziaukas, line 1547: Multiple versions supported by the action, see https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-multiple-pythonpypy-version

flowzone.yml
```diff
@@ -1529,87 +1535,39 @@ jobs:
...
-        run: |
-          error_check=`(poetry env use 3.10 2>&1 || true)`
-          if ! grep -q "Please choose a compatible version" <<< $error_check
-          then
-            echo "PYTHON_VERSIONS=$(echo "${PYTHON_VERSIONS}" | jq -c '. + ["3.10"]')" >> $GITHUB_ENV
-          else
-            echo "Python 3.10 does not meet project requirements."
-          fi
+          versions=()
+          while IFS= read -r version; do
```
pranasziaukas, line 1553: Less elegant than
```bash
for version in ${{ env.SUPPORTED_VERSIONS }}; do
```
but that's the price to pay for reusing the same configuration value in several places.

flowzone.yml
```diff
@@ -1529,87 +1535,39 @@ jobs:
...
+            echo "Setting up Python $version"
+            error_check=`(poetry env use $version 2>&1 || true)`
+            if ! grep -q "Please choose a compatible version" <<< $error_check; then
+              versions+=("$version")
+            else
+              echo "Python $version does not meet project requirements."
+            fi
+          done <<< "$(echo -n "${SUPPORTED_VERSIONS}")"
```
pranasziaukas, line 1561:
```bash
done <<< "${{ env.SUPPORTED_VERSIONS }}"
```
wouldn't work because of the trailing newline.

flowzone.yml
```diff
@@ -1529,87 +1535,39 @@ jobs:
...
+          echo "PYTHON_VERSIONS=[$(IFS=,; echo "${versions[*]}")]" >> $GITHUB_ENV
 
-      # default to the latest version (^3.7) on the runner
```
pranasziaukas, line 1606: I'd prefer to not mention a specific version here.

flowzone.yml
```diff
@@ -1529,87 +1535,39 @@ jobs:
...
+      # default to the latest version on the runner
       # if none were matched by the checks above
-      - name: Set Python versions
+      - name: Output compatible Python versions
         if: steps.python_poetry.outputs.enabled == 'true'
         id: python_versions
         run: |
-          echo "json=[\"\^3.7\"]" >> $GITHUB_OUTPUT
+          echo "json=[\"3.x\"]" >> $GITHUB_OUTPUT
```
pranasziaukas, line 1570: From the official example of specifying the latest stable version of Python (for specified major version), see https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-a-python-version

flowzone.yml
pranasziaukas: Supported (released) versions are selected, as per https://devguide.python.org/versions

flowzone.yml
```diff
@@ -1529,87 +1535,39 @@ jobs:
...
+          while IFS= read -r version; do
+            echo "Setting up Python $version"
+            error_check=`(poetry env use $version 2>&1 || true)`
+            if ! grep -q "Please choose a compatible version" <<< $error_check; then
+              versions+=("$version")
+            else
+              echo "Python $version does not meet project requirements."
+            fi
+          done <<< "$(echo -n "${SUPPORTED_VERSIONS}")"
+          echo "PYTHON_VERSIONS=[$(IFS=,; echo "${versions[*]}")]" >> $GITHUB_ENV
```
pranasziaukas, line 1562: Let's break this down for clarify:
- `IFS=,` sets the Internal Field Separator to a comma for the subsequent command
- `echo "${versions[*]}"` prints all items in the array, separated by the IFS
- `echo "PYTHON_VERSIONS=[$(...)]"` puts the comma-separated string inside square brackets to form a valid JSON array
- `>> $GITHUB_ENV` appends the result to the GitHub environment, making the `PYTHON_VERSIONS` variable available to subsequent steps

flowzone.yml
gantonayde: Yes, but look at how much duplicate code this saves. It is super worth it

flowzone.yml
gantonayde: This makes it very easy to update the list.

gantonayde (approved):
LGTM. Very solid update and much needed as it makes it very easy to maintain the support list of python versions.

PR #719, opened 2023-08-23
Always check for the presence of custom actions

klutchell:
If custom actions are not allowed, the workflow will be rejected by the custom actions job(s) and fail. With this condition in place it was silently skipping custom actions.

Change-type: patch


klutchell:
I self-certify!

Page- (approved):

PR #720, opened 2023-08-23
Do not toggle automerge for PRs authored by github apps

klutchell:
GitHub apps should have their own automerge policies, like Renovate.

Change-type: patch


klutchell:
I self-certify!

PR #730, opened 2023-08-28
Limit the github installation token to current repo only

klutchell:
This is much more secure than the default of access to all repos where the app is installed.

Change-type: minor


klutchell:
I self-certify!

PR #731, opened 2023-08-29
Switch to latest skopeo testing image

klutchell:
The stable images have been deleted, possibly by mistake, but this is blocking merges.

https://quay.io/repository/containers/skopeo?tab=history

We should install the skopeo binary in our self-hosted images going forward.

Change-type: patch


klutchell:
I self-certify!

PR #737, opened 2023-09-04
Use `docker compose run` for testing

pipex:
From docker compose documentation:

> The `docker compose run` command is for running "one-off" or "adhoc" tasks. It requires the service name you want to run and only starts containers for services that the running service depends on. Use run to run tests or perform an administrative task such as removing or adding data to a data volume container. The `run` command acts like `docker run -ti` in that it opens an interactive terminal to the container and returns an exit status matching the exit status of the process in the container.

Using `docker compose up` may lead to a false negative test if one of the other services (i.e. not the `sut` service), terminate before the service gets a chance to finish.

Change-type: patch


pipex:
I self-certify!

PR #745, opened 2023-09-11
Only setup poetry if python setup was successful

klutchell:
Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -469,8 +469,16 @@
     with:
       version: v0.14.0
 
+  - &setupPython # https://github.com/actions/setup-python
+    name: Setup python
+    id: setup-python
+    uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4
+    with:
+      python-version: "3.9"
+
   - &setupPoetry # https://github.com/abatilo/actions-poetry
     name: Setup poetry
+    if: steps.setup-python.outputs.python-version != ''
     uses: abatilo/actions-poetry@192395c0d10c082a7c62294ab5d9a9de40e48974 # v2.3.0
     with:
       poetry-version: "1.5.1"
@@ -1554,9 +1562,7 @@ jobs:
             echo "pypi_publish=false" >> $GITHUB_OUTPUT
           fi
 
-      - name: Set up Python environment
-        if: steps.python_poetry.outputs.enabled == 'true'
-        uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4
+      - <<: *setupPython
         with:
           python-version: ${{ env.SUPPORTED_VERSIONS }}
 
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1560,7 +1560,8 @@ jobs:
         with:
           python-version: ${{ env.SUPPORTED_VERSIONS }}
 
-      - *setupPoetry
+      - <<: *setupPoetry
+        if: steps.python_poetry.outputs.enabled == 'true'
```
ab77, line 1564: @klutchell why not just add it to the anchor? Assuming setup-python is always a dependency (I can't think of a reason why it wouldn't be):

```
  - &setupPoetry # https://github.com/abatilo/actions-poetry
    name: Setup poetry
    uses: abatilo/actions-poetry@192395c0d10c082a7c62294ab5d9a9de40e48974 # v2.3.0
    if: steps.python_poetry.outputs.enabled == 'true'
    with:
      poetry-version: "1.5.1"
```

klutchell: because this step `steps.python_poetry.outputs.enabled` only runs during `is_python` and doesn't exist for the other jobs that call setupPoetry

klutchell: based on the changes from @pranasziaukas , a poetry file is required for all of the downstream python jobs, it's only during `is_python` that we need to check if we found a poetry file or not

klutchell: earlier comments withdrawn, that would be a better approach

ab77 (approved):

flowzone.yml
```diff
@@ -469,8 +469,16 @@
     with:
       version: v0.14.0
 
+  - &setupPython # https://github.com/actions/setup-python
```
pranasziaukas, line 472: TIL YAML anchors

flowzone.yml
```diff
@@ -469,8 +469,16 @@
...
+    name: Setup python
+    id: setup-python
+    uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4
```
pranasziaukas, line 475: Why not 
```suggestion
    uses: actions/setup-python@v4
```

I see specific commits referenced throughout the config, is it our convention to minimize the risk of breaking things?

flowzone.yml
```diff
@@ -469,8 +469,16 @@
...
+    with:
+      python-version: "3.9"
```
pranasziaukas, line 477: Just wondering, who is to decide the appropriate version here? Do we go with the latest that is known to work?

flowzone.yml
```diff
@@ -3878,9 +3875,7 @@ jobs:
       github.event.repository.private != true
 
     steps:
-      # setup python 3.11
-      - name: Set up Python 3.11
-        uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4
+      - <<: *setupPython
         with:
           python-version: 3.11
```
pranasziaukas, line 3880: Nice to see this version override working as expected!

flowzone.yml
klutchell: Yes, Renovate will manage these digests automatically for us but the goal is to reduce the risk of introducing changes without a commit record.

flowzone.yml
klutchell: This default value is only used by `python_publish` and `python_finalize` and I'm not sure what effect a new version would have. IMO it would be best to keep it relatively recent but I didn't want to introduce a change like that in this PR.

pranasziaukas: Makes sense, thanks.

PR #759, opened 2023-09-13
Revert "Fall back to curl if github CLI fails to update objects"

klutchell:
This reverts commit 516501659b5fbbc389705f7ce33069ae8d7acbac.

Change-type: patch


klutchell:
I self-certify!

PR #762, opened 2023-09-20
Restore access to all repositories when cloning sources

klutchell:
Otherwise we lose access to private submodules on clone.

See: https://github.com/product-os/flowzone/pull/730
Change-type: patch


klutchell:
I self-certify!

PR #782, opened 2023-10-30
Add pull request read permissions and bump versionist

pipex:
Read permissions on pull requests is necessary for changelog generation on internal or private repositories

Change-type: patch


klutchell:
LGTM

PR #787, opened 2023-10-31
Reset the .github folder to the event commit for custom actions

klutchell:
The github.sha value is the tip of BASE for external PRs, and the merge commit for internal PRs so it is ideal from a security standpoint to run custom actions.

Change-type: minor


klutchell:
I self-certify!

PR #793, opened 2023-11-02
AWS/ECR public/private support

ab77:
Allows pushing images to AWS/ECR (e.g.)

* https://github.com/belodetek/cfn-generic-custom-resource/blob/master/.github/workflows/flowzone.yml#L58-L59
* https://github.com/belodetek/cfn-generic-custom-resource/pull/10


klutchell:
LGTM

PR #797, opened 2023-11-07
Fetch and checkout merge ref when resetting github directory

klutchell:
Don't assume that github.sha is the same as github.ref or even
exists in the clone. Instead fetch and checkout github.ref directly.

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -522,22 +522,22 @@ jobs:
               "metadata": "read",
               "pull_requests": "read"
             }
-      - name: Checkout merge ref
-        if: github.event.pull_request.state == 'open'
+      - name: Checkout refs/pull/${{ github.event.number }}/merge
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
         with:
           fetch-depth: 0
           submodules: recursive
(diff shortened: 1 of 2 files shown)


klutchell:
For example in this job: https://github.com/product-os/environment-staging/actions/runs/6785910945/job/18450843225

The event context contains the following:
```json
  "job": "event_types",
  "ref": "refs/pull/782/merge",
  "sha": "52fbb8ad5ebbae5b23b9f603da85514e559befd4",
  "repository": "product-os/environment-staging",
```

And for some reason the `ref` and the `sha` do not resolve to the same commit blobs.
- `GITHUB_SHA`: https://github.com/product-os/environment-staging/commit/52fbb8ad5ebbae5b23b9f603da85514e559befd4
- `GITHUB_REF`: https://github.com/product-os/environment-staging/commit/bfa1dfa28a2d1727f36a86882ba2d56ca9d441c2

Even though [this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) definition suggests they should be the same?

> `GITHUB_SHA`: Last merge commit on the `GITHUB_REF` branch
> `GITHUB_REF`: PR merge branch `refs/pull/:prNumber/merge`

So for whatever reason if the SHA is not present in the GITHUB_REF branch the checkout will fail.

PR #799, opened 2023-11-10
Always publish GitHub releases even without artifacts

klutchell:
Change-type: minor


klutchell:
@ab77 @Page- your thoughts on this change, any downsides?

The use case being https://github.com/balena-io/deploy-to-balena-action/issues/104

ab77:
> @ab77 @Page- your thoughts on this change, any downsides?
> 
> The use case being [balena-io/deploy-to-balena-action#104](https://github.com/balena-io/deploy-to-balena-action/issues/104)

I think in absence of any files, GH adds the default archive of the commit to the release, so this is fine.

PR #808, opened 2023-11-15
Look for .zst artifacts when uploading

pipex:
The job would previously look for .tar.gz artifacts which are no longer generated since switching compression

Change-type: patch


pipex:
I self-certify!

PR #809, opened 2023-11-16
Remove GitHub token secrets from docker tests

klutchell:
These were always a security concern and now there
appears to be no requirement for them.

The workaround for providing GitHub tokens to compose
tests is to encode a PAT as part of COMPOSE_VARS which
are only injected for private repositories.

Change-type: minor


klutchell:
I self-certify!

PR #810, opened 2023-11-16
Create empty .env file if no COMPOSE_VARS exist

pipex:
This should prevent failing docker compose tests due to a missing env file.

See: docker/compose#10995
Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -2309,6 +2309,7 @@ jobs:
             fi
           done
 
+          touch ${COMPOSE_ENV_FILE}
           docker compose --env-file="${COMPOSE_ENV_FILE}" --project-directory="$(pwd)" ${args} config > "${COMPOSE_FILE}"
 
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -2280,6 +2280,8 @@ jobs:
               secret="$(echo "${line}" | awk -F'=' '{print $2}')"
               echo "::add-mask::${secret}"
             done < ${COMPOSE_ENV_FILE}
+          else
+            touch ${COMPOSE_ENV_FILE}
```
klutchell, line 2284: This step only runs for internal PRs, so it will still fail for external PRs.

klutchell: move the creation to the next step and it should be fine

flowzone.yml
```diff
@@ -2309,6 +2309,7 @@ jobs:
             fi
           done
 
+          touch ${COMPOSE_ENV_FILE}
```
pipex, line 2312: @klutchell is this what you meant?

klutchell (approved):

PR #818, opened 2023-11-28
Use buildx as a more reliable check of platform support

klutchell:
When running binfmt in a guest docker daemon on arm64 the native support for arm/v6 and arm/v7 is not reflected.

However buildx detects support for those platforms just fine.

Change-type: patch


klutchell:
I self-certify!

PR #820, opened 2023-11-30
Fix native platform check before registering QEMU binfmt

klutchell:
This avoids matching substrings of the natively supported platforms by converting the list to json first.

Change-type: patch


klutchell:
I self-certify!

Issue #729, opened 2023-08-28
Repo description with a single quote in the name causes an error


gantonayde:
Post-merge flowzone action failed during `Apply repo settings/Configure repository` step https://github.com/people-os/support-shift-scheduler/actions/runs/6000652245/job/16273493272#step:3:57

with:

```bash
Run # only change repository visibility if explicitly set to one of the permissible values
  # only change repository visibility if explicitly set to one of the permissible values
  visibility=''
  if [[ 'default' =~ private|public|internal ]]; then
      gh repo edit 'people-os/support-shift-scheduler' \
        --visibility 'default' || true
  fi
  
  if [[ -n 'Project for scheduling Balena's engineers for customer support' ]]; then
      gh repo edit 'people-os/support-shift-scheduler' \
        --description 'Project for scheduling Balena's engineers for customer support'
  fi
  
  if [[ -n '' ]]; then
      homepage=''
  elif [[ -n '' ]]; then
      homepage='https://.pages.dev'
  else
      homepage=''
  fi
  [[ -n "${homepage}" ]] && gh repo edit 'people-os/support-shift-scheduler' \
    --homepage "${homepage}"
  
  # HTTP 4[2](https://github.com/people-os/support-shift-scheduler/actions/runs/6000652245/job/16273493272#step:3:2)2: This organization does not allow private repository forking
  if ! gh repo edit 'people-os/support-shift-scheduler' \
    --allow-forking=true; then
      echo '::warning::Failed to configure some repository settings.'
  fi
  
  # FIXME: https://github.com/cli/cli/issues/6652#issuecomment-1[3](https://github.com/people-os/support-shift-scheduler/actions/runs/6000652245/job/16273493272#step:3:3)23908232
  gh repo edit 'people-os/support-shift-scheduler' \
    --default-branch=master \
    --delete-branch-on-merge=true \
    --enable-auto-merge=true \
    --enable-issues=true \
    --enable-merge-commit=true \
    --enable-projects=false \
    --enable-rebase-merge=false \
    --enable-squash-merge=false \
    --enable-wiki=false
  
  # allow_update_branch not currently available for update via gh-cli
  gh api --method PATCH '/repos/people-os/support-shift-scheduler' \
    -H "Accept: application/vnd.github+json" \
    -F allow_update_branch='true'
  shell: /usr/bin/bash -e {0}
  env:
    NPM_REGISTRY: https://registry.npmjs.org
    CARGO_REGISTRY: crates.io
[...]

PR #833, opened 2023-12-19
BREAKING: Remove opt-in repo configuration

klutchell:
Allowing Flowzone to make repo configuration changes
via administrator tokens effectively grants any user with
write permissions the ability to make admin changes.

Repo configuration should be handled outside of the CI
context where administrator permissions can be granted to
a central authority only.

This is a breaking change for repos that previously opted into these settings.
This workflow will not run unless the calling workflow omits the related inputs.

Change-type: major
See: https://balena.fibery.io/Work/Improvement/Move-repo-config-out-of-Flowzone-CI-1388
See: https://balena.fibery.io/Work/Project/Host-the-safe-settings-GitHub-App-for-repository-management-190

Resolves: https://github.com/product-os/flowzone/issues/729


klutchell:
62 repositories appear to be affected, some of them external to balena orgs.

ab77 (approved):

PR #882, opened 2024-02-07
Publish rust binaries in tar.gz format

pipex:
This reverts a previous move to tar.zst. While zstd provides better
compression, tar.gz is a more commonly used format and is compatible
accross multiple platforms

This is a major change as it changes the expected output from an
existing process.

Change-type: major


klutchell:
LGTM

Issue #878, opened 2024-01-30
Warning: Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20


klutchell:
```
Flowzone / Is rust
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Flowzone / Is custom
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Flowzone / Is docker
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Flowzone / Test docker (balena-push-env, linux/amd64, ubuntu-22.04)
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Flowzone / Test docker (block-build-env, linux/amd64, ubuntu-22.04)
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Flowzone / Test docker (default, linux/amd64, ubuntu-22.04)
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Flowzone / Test docker (generate-ami-env, linux/amd64, ubuntu-22.04)
[...]

PR #896, opened 2024-02-16
Use bash to build json arrays and drop unmaintained action

klutchell:
Change-type: patch
See: https://github.com/product-os/flowzone/issues/878

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -272,7 +272,14 @@
       INPUT="$(echo "${INPUT}" | tr -d '[:space:]')"
 
       # Convert to JSON array using jq to split the string by the given delimiter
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -262,12 +262,26 @@
...
-    uses: kanga333/json-array-builder@c7cd9d3a8b17cd368e9c2210bc3c16b0e2714ce5 # v0.2.1
-    with:
-      cmd: bash -c "echo $INPUT | tr -d '[:space:]'"
-      separator: ","
+    shell: bash
+    run: |
+      set -x
+
+      # Remove spaces and newlines from the input
+      INPUT="$(echo "${INPUT}" | tr -d '[:space:]')"
```
Page-, line 272: I think this is deleting all spaces, not just those around the quoted elements, but it's also not new to this PR

Page- (approved):

PR #946, opened 2024-03-19
use balena cli to set release notes

ab77:
This is ready to go.

image

image

* https://github.com/balena-io/deploy-to-balena-action/issues/194
* https://github.com/balena-io/deploy-to-balena-action/pull/341

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -35,7 +35,7 @@
     uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
     # skip if the PR is from a fork or aws region is unset
     if: |
-      ( matrix.region != '' || inputs.aws_region != '' ) && 
+      ( matrix.region != '' || inputs.aws_region != '' ) &&
       github.event.pull_request.head.repo.full_name == github.repository
     continue-on-error: true
     with:
@@ -67,6 +67,7 @@
       environment: ${{ inputs.balena_environment }}
       fleet: ${{ matrix.slug }}
       source: ${{ inputs.working_directory }}
+      note: ${{ steps.release_notes.outputs.note }}
       registry_secrets: |
         {
           "ghcr.io": {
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -67,6 +67,7 @@
       environment: ${{ inputs.balena_environment }}
       fleet: ${{ matrix.slug }}
       source: ${{ inputs.working_directory }}
+      note: ${{ steps.release_notes.outputs.note }}
```
ab77, line 70: the only side effect I found is the addition of extra quotes around the note:

image

flowzone.yml
```diff
@@ -86,30 +87,22 @@
...
-      release_id='${{ steps.balena_deploy.outputs.release_id }}'
-      if [[ -n $release_notes ]] && [[ -n $app_id ]] && [[ -n $release_id ]]; then
-          curl --silent --retry 3 --fail \
-            -X PATCH "https://api.${{ inputs.balena_environment }}/v6/release?\$filter=belongs_to__application%20eq%20${app_id}%20and%20id%20eq%20${release_id}" \
-            -H 'Content-Type: application/json' \
-            -H 'Authorization: Bearer ${{ secrets.BALENA_API_KEY || secrets.BALENA_API_KEY_PUSH }}' \
-            -d "{\"note\":${release_notes}}"
+      look_back=1
+      if [[ '${{ github.event.action }}' =~ closed ]] && [[ '${{ github.event.pull_request.merged }}' == true ]]; then
+          look_back=2
```
ab77, line 93: this additional logic should select the correct range depending whether the release(s) are draft or finalised.

flowzone.yml
klutchell: is that coming from the DtB action or Flowzone? We should identify the fix as it's pretty ugly IMO

flowzone.yml
klutchell: I don't think we can hardcode that logic in this function as the caller is more aware of what is required to find the last tag for a diff.

For example, the `balena_publish` job (and others) are calling `*createLocalRefs` so the draft tag actually exists locally in those cases. It should behave similar to merged and look back 2 refs, shouldn't it?

What if we make `look_back: 2` an env var on this step, and calling jobs can override it if necessary?

flowzone.yml
```diff
@@ -2481,9 +2473,8 @@ jobs:
     steps:
       - *getGitHubAppToken
       - *checkoutVersionedSha
-      - *deployToBalenaAction
       - *getReleaseNotes
```
klutchell, line 2476:
```suggestion
      - << *getReleaseNotes
        env:
          look_back: 2
```
example of providing a custom look_back value from the calling job

flowzone.yml
ab77: https://github.com/balena-io/deploy-to-balena-action/pull/341

klutchell (approved):

PR #947, opened 2024-03-19
Add GitHub environments support to CFN workflows

ab77:


flowzone.yml
```diff
@@ -3290,6 +3318,19 @@ jobs:
 
           echo "caps=${caps}" >> $GITHUB_OUTPUT
 
+      - name: Install cfn-lint
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install cfn-lint==0.83.7
+
+      - name: Lint template
```
klutchell, line 3326: Is this just a linter for CF templates? I would prefer to review this in a separate PR from the environment stuff.

flowzone.yml
```diff
@@ -2998,6 +3006,9 @@ jobs:
       matrix:
         value: ${{ fromJSON(needs.is_custom.outputs.custom_test_matrix) }}
         os: ${{ fromJSON(inputs.custom_runs_on || format('[{0}]', inputs.runs_on)) }}
+        include: ${{ fromJSON(inputs.custom_matrices_includes) }}
+
+    environment: ${{ matrix.environment || null }}
```
klutchell, line 3011: I think environment(s) should be fully supported by Flowzone at the top level and not require custom matrices, even though this looks like it would solve the immediate problem at hand.

Can we brainstorm about a more generic approach next week?

flowzone.yml
ab77: in that case, we should do that one first as it prevents garbage from being committed.

klutchell:
Any thoughts on my additions @ab77 ?

I've tested it [here](https://github.com/balena-io/environment-playground/pull/69) but I'm not sure if there is anything else to look for except that custom tests are passing.

klutchell:
LGTM

PR #950, opened 2024-03-21
update production floating tag

ab77:
change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -102,6 +102,48 @@
       echo "${release_notes}" > "${release_notes_file}"
       echo "file=${release_notes_file}" >> $GITHUB_OUTPUT
 
+  - &getTimeStamp
+    name: Get and format timestamp
+    id: timestamp
+    env:
+      FORMAT: '%Y-%m-%d-%H%M%S'
+    run: |
+      echo "datetime=$(date +"${FORMAT}")" >> $GITHUB_OUTPUT
+
+  # https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#update-a-reference
+  - &updateGitReference
+    name: Update git reference
+    if: github.event.pull_request.merged == true && steps.create_commit.outputs.sha != ''
+    env:
+      GH_TOKEN: ${{ steps.gh_app_token.outputs.token || secrets.FLOWZONE_TOKEN }}
+      REF: '/repos/$GH_REPO/git/refs/heads/${{ github.base_ref }}'
+    run: |
+      gh api \
+        -X PATCH \
+        -H "Accept: application/vnd.github+json" \
+        -H "X-GitHub-Api-Version: 2022-11-28" \
+        "${REF}" \
+        -f sha='${{ steps.create_commit.outputs.sha }}' \
+        -F force=true \
+        --include
+
+  # https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#create-a-reference
+  - &createGitReference
+    name: Create git reference
+    if: github.event.pull_request.merged == true && steps.create_tag.outputs.sha != ''
+    env:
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1180,34 +1180,45 @@ jobs:
           echo "json=$(echo $response | jq -c .)" >> $GITHUB_OUTPUT
 
       # https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#update-a-reference
-      - name: Update branch reference
+      - name: Update git references
         if: github.event.pull_request.merged == true && steps.create_commit.outputs.sha != ''
+        # some tags may not initially exist (e.g. production)
+        continue-on-error: true
```
klutchell, line 1186: This should be a separate step so we don't ignore errors when updating the branch ref.

flowzone.yml
```diff
@@ -1180,34 +1180,45 @@ jobs:
...
-            -X PATCH \
-            -H "Accept: application/vnd.github+json" \
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            /repos/$GH_REPO/git/refs/heads/${{ github.base_ref }} \
-            -f sha='${{ steps.create_commit.outputs.sha }}' \
-            -F force=true \
-            --include
+          for ref in \
+            '/repos/$GH_REPO/git/refs/heads/${{ github.base_ref }}' \
+            '/repos/$GH_REPO/git/refs/tags/production'; do
```
klutchell, line 1192: The production tag seems arbitrary, especially without a commit message or comment to explain where it is used.

Maybe we should create an env or input for floating tags that need to be updated on merge rather than hardcoding something that is only used for a few internal repos.

Then we wouldn't need to ignore failures if the floating tag is provided by the calling repo.

klutchell (changes requested):

flowzone.yml
ab77: `balena-deploy-request`, which is used to generate release notes, requires this tag. It should probably be set as a default to `production`, with a way to override it (though I am struggling to see a use case for the later).

Ultimately we want this in as a prerequisite to Flowzone versioning to add a PR comment to the PR for the approver to adjust (notable changes) prior to merging.

klutchell: Yup, I understand the use case but no one in the future looking at this code will know why it is there and why it fails on most repos except some balena backend components.

It just seems like something that should be opt-in.

ab77: OK, how do you propose this opt-in to work? Another input (e.g. `generate_release_notes`)?

klutchell: An input like floating_tags might do it with a good enough description.

ab77: `balena-deploy-request` requires minimum of two tags in the format `production-$(date +%Y-%m-%d-%H%M%S)`, so the input would have to be something like `floating_tag_prefix` or just a boolean `floating_tags`, or some sort of weird template format, which we would then have to parse and substitute env vars, etc.

I suggest a simple boolean input called `release_notes`, which will toggle the entire workflow (production tags and PR comment).

flowzone.yml
klutchell: Ignoring errors does not hide them from the job summary. This step will fail on 99% of repos and leave errors in the summary (even when the build continues).

I think this should be opt-in or we should make it possible to suppress the error only for the production tags.

Maybe we should add another env var to optionally ignore errors?

Something like `gh api ... ${{ env.APPEND_CMD }}` where the default is `` but for production we pass `|| exit 0`?

flowzone.yml
ab77: @klutchell thoughts, prayers?

klutchell: > I suggest a simple boolean input called `release_notes`, which will toggle the entire workflow (production tags and PR comment).

This is probably the best approach, go for it.

PR #953, opened 2024-03-25
run steps on X64 only

ab77:
* https://github.com/actions/setup-python/issues/705
* https://github.com/balena-io/environment-production/actions/runs/8423822603/job/23066477310?pr=2178#step:8:24

change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -3216,11 +3216,6 @@ jobs:
       - *getAWSCallerIdentity
       - *convenienceFunctions
 
-      - <<: *setupPython
-        with:
-          cache: 'pip' # caching pip dependencies
-          python-version: 3.11
-
       - name: Create templates bucket
         id: make_bucket
         run: |
@@ -3291,8 +3286,7 @@ jobs:
 
       - name: Install cfn-lint
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install cfn-lint==0.83.7
+          python3 -m pip install cfn-lint==0.83.7
 
       - name: Lint template
         run: |
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -3732,12 +3732,6 @@ jobs:
           }
           $EOF
           echo "with_backoff=${with_backoff}" >> $GITHUB_OUTPUT
-      - name: Setup python
-        id: setup-python
-        uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c
-        with:
-          cache: pip
-          python-version: 3.11
       - name: Create templates bucket
         id: make_bucket
         run: |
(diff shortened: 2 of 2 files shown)


flowzone.yml
```diff
@@ -3216,6 +3217,7 @@ jobs:
       - *getAWSCallerIdentity
       - *convenienceFunctions
 
+      # FIXME: https://github.com/actions/setup-python/issues/705
```
klutchell, line 3220: We use setup python all over this workflow, why is this the first time we encountered issues with it on some runners?

Which specific runners are the issue? Can't we just use the python version already installed on the runner?

flowzone.yml
```diff
@@ -3184,7 +3184,8 @@ jobs:
 
   cloudformation_test:
     name: Test CloudFormation
-    runs-on: ${{ fromJSON(inputs.cloudformation_runs_on || inputs.runs_on) }}
+    # FIXME: https://github.com/actions/setup-python/issues/705
+    runs-on: ["ubuntu-22.04", "X64"]  # ${{ fromJSON(inputs.cloudformation_runs_on || inputs.runs_on) }}
```
klutchell, line 3188: This tag combination isn't valid, it will look for a runner with both `ubuntu-22.04` (GitHub hosted) and `X64` which AFAIK only exists on self-hosted.

IMO we shouldn't be hardcoding runners, but instead should skip steps when incompatible runners are detected.

klutchell (changes requested):

flowzone.yml
ab77: Python only runs fro two types on projects: CloudFormation and Python. The former is triggered for any environment repo with AWS templates, the later, I have no idea - I don't think we have many active Python projects.

klutchell: Can we just use the existing python version on the runner to save a step and possible breakages? It's just a linter after all, it shouldn't need latest python or latest pip and those just add time and complexity.

ab77: We can, but this is going wreck havoc with other Python dependencies (since we aren't setting up venvs). Might just work enough though.

klutchell: The runner is destroyed after the job anyway, so any havoc would be contained to the CF test job at least.

klutchell (approved):

ab77:
This broke AWS-CLI: https://github.com/balena-io/environment-playground/actions/runs/8425936303/job/23073177694?pr=62#step:13:46

klutchell:
Dang, but we don't want to hardcode Flowzone jobs to a particular runner, especially if the calling repo is only supposed to only use self-hosted, or the other way around.

We can revert this commit and filter by runner architecture when deciding whether or not to setup python and perform the linting.

eg. `if: runner.arch == X64` on the steps related to this python package?

PR #955, opened 2024-03-25
revert product-os/flowzone/pull/953

ab77:
* installing Python dependencies on runner [breaks](https://github.com/balena-io/environment-playground/actions/runs/8425936303/job/23073177694?pr=62#step:13:38) other dependencies (e.g. AWS-CLI)

change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -3284,9 +3284,15 @@ jobs:
 
           echo "caps=${caps}" >> $GITHUB_OUTPUT
 
+      - <<: *setupPython
+        with:
+          cache: 'pip' # caching pip dependencies
+          python-version: 3.11
+
       - name: Install cfn-lint
         run: |
-          python3 -m pip install cfn-lint==0.83.7
+          python -m pip install --upgrade pip
+          python -m pip install cfn-lint==0.83.7
 
       - name: Lint template
         run: |
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -3818,9 +3818,16 @@ jobs:
           fi
 
           echo "caps=${caps}" >> $GITHUB_OUTPUT
+      - name: Setup python
+        id: setup-python
+        uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c
+        with:
+          cache: pip
+          python-version: 3.11
       - name: Install cfn-lint
         run: |
-          python3 -m pip install cfn-lint==0.83.7
+          python -m pip install --upgrade pip
+          python -m pip install cfn-lint==0.83.7
       - name: Lint template
         run: |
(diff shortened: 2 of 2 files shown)


klutchell:
I prefer [this](https://github.com/product-os/flowzone/pull/953#issuecomment-2018787249) approach rather than hardcoding runner tags.

ab77:
> I prefer [this](https://github.com/product-os/flowzone/pull/953#issuecomment-2018787249) approach rather than hardcoding runner tags.

How is this approach fix the underlying condition that the setup-python action will **only** run on X64/Linux currently?

flowzone.yml
```diff
@@ -3184,7 +3184,8 @@ jobs:
 
   cloudformation_test:
     name: Test CloudFormation
-    runs-on: ${{ fromJSON(inputs.cloudformation_runs_on || inputs.runs_on) }}
+    # FIXME: https://github.com/actions/setup-python/issues/705
+    runs-on: ["Linux", "X64"] #${{ fromJSON(inputs.cloudformation_runs_on || inputs.runs_on) }}
```
klutchell, line 3188: This is forcing self-hosted runners, which are not available to all repositories, especially outside of our org.

Also it doesn't address the other jobs that are using setup-python.

klutchell:
> > I prefer [this](https://github.com/product-os/flowzone/pull/953#issuecomment-2018787249) approach rather than hardcoding runner tags.
> 
> How is this approach fixing the underlying condition that the setup-python action will **only** currently succeed on X64/Linux? I must be missing something, perhaps you can add you commit directly to this branch?
> 
> We have seven environment repositories that are affected by this. We could technically change them all to only run on X64/Linux, but I think a one line change here is better than having to change x7, especially given that it will be temporary.

HOW temporary? Do we know if/when this will be fixed upstream?

I see the benefit of fixing it in Flowzone as it's less work, but then we have to make assumptions about what the calling repository has access to (self-hosted vs GitHub hosted) AND we need to remember and come back later to remove all the places in Flowzone where we are ignoring the `runs_on` input to provide a hardcoded one.

Would [this](https://github.com/actions/setup-python/issues/705#issuecomment-1756948951) be a suitable compromise?

ab77:
> Would [this](https://github.com/actions/setup-python/issues/705#issuecomment-1756948951) be a suitable compromise?

Ugh..

flowzone.yml
ab77: I am tempted to leave those to the project maintainer(s).

ab77:
> > Would [this](https://github.com/actions/setup-python/issues/705#issuecomment-1756948951) be a suitable compromise?
> 
> Ugh..

We'll have to do it [upstream](https://github.com/balena-io/environment-playground/pull/62) and revert when they get their sh*t together.

klutchell (approved):

PR #957, opened 2024-03-26
adds release notes into PR as a comment

ab77:
uses Renovate's release notes from the PR description (e.g.) throw-away/superseded PR

* https://github.com/balena-io/balena-cloud/pull/2907#issuecomment-2073662535
* Zulip messages (on merge) remain to be tested, but *should* work

.. duplicates need to be fixed [upstream](https://github.com/renovatebot/renovate/issues/13037)

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ jobs:
       # Required: false
       PYPI_TEST_TOKEN:
 
+      # API key to post Zulip messages.
+      # Required: false
+      ZULIP_API_KEY:
+
       # Optional secret for using with custom jobs
       # Required: false
       CUSTOM_JOB_SECRET_1:
@@ -372,6 +376,11 @@ jobs:
       # Type: string
       # Required: false
       ok_to_test_label: ok-to-test
+
+      # Create git tags and a PR comment with detailed change log.
+      # Type: boolean
+      # Required: false
+      release_notes: false
 
 
 ```
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -81,7 +81,7 @@
         }
 
   - &getReleaseNotes
-    name: Generate release notes
+    name: Generate short release note
     id: release_notes
     env:
       look_back: 1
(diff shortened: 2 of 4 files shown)


flowzone.yml
```diff
@@ -1032,12 +1092,13 @@ jobs:
...
           node-version: 18.x
 
-      - name: Install versionist
+      - name: Install versioning tools
         if: inputs.disable_versioning != true
         run: |
-          npm install -g balena-versionist@~0.14.13 versionist@^7.0.3
-          npm ls -g balena-versionist
-          npm ls -g versionist
+          cat <package.json | jq -r '.dependencies | to_entries[] | .key + "@" + .value' \
```
Page-, line 1098: This would be looking at the `package.json` of the repo we're building, *not* of flowzone, so this won't work as expected outside of the flowzone repo

klutchell (changes requested):
Flowzone seems like the wrong place for this, especially with all the Zulip requirements. Could we not just create a shared reusable workflow similar to Flowzone and trigger it from the respective projects on pull_request?

Also, if we are going to automate the Zulip messages on PR merge, why not use the same workflow for the manual deploy buttons to consolidate the workflows?

ab77:
**tl;dr** I strongly advocate for a single workflow for both deploys and regular updates.

> Flowzone seems like the wrong place for this, especially with all the Zulip requirements. Could we not just create a shared reusable workflow similar to Flowzone and trigger it from the respective projects on pull_request?

Yes we can, however that would defeat the main purpose of this work, which is to bring the "deploy process" under a single umbrella. If we accept Renovate PRs as the source of truth for any change, then it only makes sense to keep all deployment activities within the scope of these PRs.

This is what these changes attempt to accomplish. Renovate already provides us with the necessary release notes/context in each update PR, so instead of manually generating these again using our own tool (which we also have to maintain) and copying them into a message, we automate it in the logical place (i.e. Renovate PR/Flowzone).

Yes we can have a separate workflow/buttons/etc., but why make it more complicated? We already have a nice working workflow, which streams updates from source to our backends - let's take advantage of this. This also removes any requirement to document and keep updated process documents on "how to do deploys", it's all documented in code and any documentation can be reduced down to a statement (i.e.) "edit PR comment, approve and merge".

> Also, if we are going to automate the Zulip messages on PR merge, why not use the same workflow for the manual deploy buttons to consolidate the workflows?

I'd like us to start moving away from the concept of "manual deploys" and instead refer to those as "break glass" events. If we are ever in a position to require a manual change, we use the "break glass" process, which would involve quiescing whatever GitOps operator we are using (e.g. Flux) and make a direct change to its state/config; or alternatively open an separate PR to make the change there, force/merging if required and generate release notes manually using `balena-deploy-request` or similar.

Have I missed anything? Thoughts, comments, prayers?


@klutchell @thgreasi @Page- @cmfcruz

flowzone.yml
```diff
@@ -907,6 +952,24 @@ on:
         type: string
         required: false
         default: "ok-to-test"
+      release_notes:
+        description: "Create git tags and a PR comment with detailed change log."
+        type: boolean
+        required: false
+        default: false
+      release_notes_zulip_stream:
```
klutchell, line 960: Can we read this from a var instead of adding a new input?

flowzone.yml
```diff
@@ -907,6 +952,24 @@ on:
...
+        description: "Zulip stream to notify."
+        type: string
+        required: false
+      release_notes_zulip_topic:
```
klutchell, line 964: Can we read this from a var instead of adding a new input?

flowzone.yml
```diff
@@ -907,6 +952,24 @@ on:
...
+        description: "Send release notes to this Zulip topic within the stream."
+        type: string
+        required: false
+      release_notes_zulip_url:
```
klutchell, line 968: Can we read this from a var instead of adding a new input?

README.md
```diff
@@ -136,6 +136,10 @@ jobs:
       # Required: false
       PYPI_TEST_TOKEN:
 
+      # API key to post Zulip messages.
+      # Required: false
+      FLOWZONE_ZULIP_API_KEY:
```
klutchell, line 141:
```suggestion
      ZULIP_API_KEY:
```

flowzone.yml
```diff
@@ -1154,35 +1220,156 @@ jobs:
...
+          <<: *getGitHubAppTokenWith
+          permissions: >-
+            {
+              "contents": "read",
+              "metadata": "read",
+              "pull_requests": "write"
+            }
+
+      # Get Renovate release notes from PR body, massage a little and draft a comment.
+      - name: Draft PR comment
```
klutchell, line 1239: Could we put the release notes, PR comment, and Zulip message in another job that depends on versioned_source? It will make it easier to track the errors and we don't have to remember to put `inputs.release_notes == true` on every step if the new job requires it.

flowzone.yml
```diff
@@ -1016,12 +1079,14 @@ jobs:
...
         if: inputs.disable_versioning != true
         run: |
-          npm install -g balena-versionist@~0.14.13 versionist@^7.0.3
-          npm ls -g balena-versionist
-          npm ls -g versionist
+          npm install -g \
+            balena-versionist@~0.14.13 \
+            versionist@^7.0.3
+
+          npm ls -g
```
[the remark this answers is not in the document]
ab77: just a refactor, no functional changes.

klutchell (approved):
Looks good to me, especially since it should be no-op until `inputs.release_notes == true` on some repos!

PR #965, opened 2024-03-28
Verify that all required docker artifacts exist

klutchell:
By setting up the publish matrix ahead of time with
the expected platforms, slugs, etc. we can process each
artifact by name rather than glob matching.


ab77 (approved):
@klutchell very comprehensive, I eye-balled it and scrolled through the publish steps.

PR #968, opened 2024-04-03
lint with actionlint

ab77:
mostly quoting env vars, but there are some other recommendations in there..

### DIFF
--- a/.github/actionlint-matcher.json
+++ b/.github/actionlint-matcher.json
@@ -1,0 +1,17 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "actionlint",
+      "pattern": [
+        {
+          "regexp": "^(?:\\x1b\\[\\d+m)?(.+?)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*: (?:\\x1b\\[\\d+m)*(.+?)(?:\\x1b\\[\\d+m)* \\[(.+?)\\]$",
+          "file": 1,
+          "line": 2,
+          "column": 3,
+          "message": 4,
+          "code": 5
+        }
+      ]
+    }
+  ]
+}
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -55,9 +55,9 @@
       decimal_sha=$(echo "ibase=16; $commit_sha" | bc)
       version_tag="${version}-dev${decimal_sha}"
 
-      echo "package=${package}" >> $GITHUB_OUTPUT
-      echo "version=${version}" >> $GITHUB_OUTPUT
-      echo "version_tag=${version_tag}" >> $GITHUB_OUTPUT
+      echo "package=${package}" >> "${GITHUB_OUTPUT}"
+      echo "version=${version}" >> "${GITHUB_OUTPUT}"
+      echo "version_tag=${version_tag}" >> "${GITHUB_OUTPUT}"
 
   - &deployToBalenaAction
     uses: balena-io/deploy-to-balena-action@3628206a0a2ecbc3a4fa717c77336545558e1447 # v2.0.49
@@ -95,12 +95,12 @@
       release_notes_file="$(mktemp)"
       release_notes="$(git log ${previous_tag}..${{ github.event.pull_request.head.sha || github.event.head_commit.id }} --pretty=reference)"
       EOF="$(openssl rand -hex 16)"
-      echo "note<<$EOF" >> $GITHUB_OUTPUT
-      echo "${release_notes}" >> $GITHUB_OUTPUT
-      echo "$EOF" >> $GITHUB_OUTPUT
+      echo "note<<$EOF" >> "${GITHUB_OUTPUT}"
+      echo "${release_notes}" >> "${GITHUB_OUTPUT}"
+      echo "$EOF" >> "${GITHUB_OUTPUT}"
 
       echo "${release_notes}" > "${release_notes_file}"
-      echo "file=${release_notes_file}" >> $GITHUB_OUTPUT
+      echo "file=${release_notes_file}" >> "${GITHUB_OUTPUT}"
 
   - &checkoutMergeRef
     # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
@@ -149,10 +149,10 @@
     id: git_describe
     run: |
       tag="$(git tag --points-at HEAD | tail -n1)"
-      echo "tag=${tag}" >> $GITHUB_OUTPUT
-      echo "semver=$(npx -q -y -- semver -c -l "${tag}")" >> $GITHUB_OUTPUT
-      echo "describe=$(git describe --tags --always --dirty | cat)" >> $GITHUB_OUTPUT
-      echo "sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+      echo "tag=${tag}" >> "${GITHUB_OUTPUT}"
+      echo "semver=$(npx -q -y -- semver -c -l "${tag}")" >> "${GITHUB_OUTPUT}"
+      echo "describe=$(git describe --tags --always --dirty | cat)" >> "${GITHUB_OUTPUT}"
+      echo "sha=$(git rev-parse HEAD)" >> "${GITHUB_OUTPUT}"
 
   - &createLocalRefs
     # Create a local reference for the versioned tag so that it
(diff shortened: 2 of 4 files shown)


.github/actionlint-matcher.json
```diff
@@ -0,0 +1,17 @@
...
+          "file": 1,
+          "line": 2,
+          "column": 3,
+          "message": 4,
+          "code": 5
+        }
+      ]
+    }
+  ]
+}
```
klutchell, line 17: Where does this file come from? Would we ever need to change it?

flowzone.yml
```diff
@@ -3061,7 +3094,7 @@ jobs:
       - *getGitHubAppToken
       - *checkoutVersionedSha
 
-      - name: Set up toolchain ${{ matrix.target }}
```
klutchell, line 3064: Any idea why it removed this?

flowzone.yml
```diff
@@ -668,6 +670,12 @@ on:
       CF_API_TOKEN:
         description: "Cloudflare API token with limited access for Pages projects"
         required: false
+      PYPI_TOKEN:
+        description: "Token to access pypi.org Python packages."
+        required: false
+      PYPI_TEST_TOKEN:
+        description: "TODO: do we need both PyPi tokens?"
+        required: false
```
klutchell, line 678: Did you add these or did the linter?

Probably best to add the lint action in one commit, and perform the lint in another commit for visibility.

klutchell (commented):
This is pretty cool, I like most of the suggestions. We should perform the lint as part of our husky pre-commit hooks.

klutchell:
> This is pretty cool, I like most of the suggestions. We should perform the lint as part of our husky pre-commit hooks.

I went and skimmed the docs and still think this would be better to run as a pre-commit hook via npm. What is the benefit of running after the changes have been pushed?

dfunckt:
The two serve different purposes. The pre-commit hook is nice for DevExp but the CI check is needed as well to catch unformatted commits. I think the latter is necessary, the former great to have.

flowzone.yml
ab77: because `${{ matrix.target }}` is not defined in this context/job.

flowzone.yml
ab77: Yeah, that was added by me, because the linter picked up the missing secrets. Perhaps you know if we need both?

ab77:
> > This is pretty cool, I like most of the suggestions. We should perform the lint as part of our husky pre-commit hooks.
> 
> I went and skimmed the docs and still think this would be better to run as a pre-commit hook via npm. What is the benefit of running after the changes have been pushed?

can you add a commit to run in a hook as well as?

klutchell:
> > > This is pretty cool, I like most of the suggestions. We should perform the lint as part of our husky pre-commit hooks.
> > 
> > 
> > I went and skimmed the docs and still think this would be better to run as a pre-commit hook via npm. What is the benefit of running after the changes have been pushed?
> 
> can you add a commit to run in a hook please?

I'll make some time next week to remove the action from the workflow and move it to an npm pre-commit hook.

.github/actionlint-matcher.json
ab77: https://github.com/rhysd/actionlint/blob/main/.github/actionlint-matcher.json

flowzone.yml
```diff
@@ -55,9 +55,9 @@
       decimal_sha=$(echo "ibase=16; $commit_sha" | bc)
       version_tag="${version}-dev${decimal_sha}"
 
-      echo "package=${package}" >> $GITHUB_OUTPUT
-      echo "version=${version}" >> $GITHUB_OUTPUT
-      echo "version_tag=${version_tag}" >> $GITHUB_OUTPUT
+      echo "package=${package}" >> "${GITHUB_OUTPUT}"
+      echo "version=${version}" >> "${GITHUB_OUTPUT}"
+      echo "version_tag=${version_tag}" >> "${GITHUB_OUTPUT}"
```
klutchell, line 60: Did you have to manually fix all of these linter errors in your editor? Did you use the VSCode plugin?
There [doesn't appear to be](https://github.com/rhysd/actionlint/issues/358) a `-fix` option for actionlint.

ab77: I don't use VSCode period. I use basic (sometimes grep) search/replace in BBEdit.

flowzone.yml
```diff
@@ -1027,6 +1035,23 @@ jobs:
         with:
           node-version: 18.x
 
+      # https://github.com/rhysd/actionlint/blob/main/docs/usage.md#problem-matchers
+      - name: Check workflow files
+        run: |
+          set -xe
+
+          echo "::add-matcher::.github/actionlint-matcher.json"
```
klutchell, line 1043: This file only exists in the Flowzone repo, so this step will always fail on all other repos.

I'll move this check to the Flowzone `tests.yml` workflow for now so we don't break all other repos.

klutchell:
Superseded by https://github.com/product-os/flowzone/pull/970

klutchell:
> shellcheck all the in-line yaml

yup, I've got a follow-up PR in the works for that, but it's likely to break workflows in other repositories so we need to be prepared to start putting out fires

we could always fix the yaml with the linter locally before enforcing it on all repositories

ab77:
> > shellcheck all the in-line yaml
> 
> yup, I've got a follow-up PR in the works for that, but it's likely to break workflows in other repositories so we need to be prepared to start putting out fires
> 
> we could always fix the yaml with the linter locally before enforcing it on all repositories

in this case it will never be enforced, especially given that `npm run lint` is entirely optional and up to the user @klutchell

klutchell:
@ab77 `npm run lint` is included in the pre-commit hooks, so unless those are being skipped intentionally it will still run locally and fail to create commits that do not pass.

If we need it enforced in the PR checks we could add `npm run lint` to the end of `npm run test` but we are going to enable shellcheck in the action soon enough anyway, so seems unnecessary.

flowzone.yml
ab77: @klutchell I must be blind, but I can't see where we are now ignoring checks for unquoted variables (npm run lint)...

klutchell: As of my [last PR](https://github.com/product-os/flowzone/pull/977/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R7), we are not ignoring checks for unquoted variables in `npm run lint`.

But shellcheck is still disabled in the action until we get an idea of how many repos will break if enforced.
https://github.com/product-os/flowzone/blob/014534c836dee687edd773ed84baea940995fc32/flowzone.yml#L1210-L1211

ab77: OK, something isn't right then:

```
$ git branch
* ab77/release-notes

$ git rebase master
Current branch ab77/release-notes is up to date.

$ actionlint .github/workflows/flowzone.yml; echo $?
0

$ cat foo.sh 
#!/usr/bin/env sh

FOO=bar

echo $foo

$ shellcheck foo.sh 

In foo.sh line 3:
FOO=bar
^-^ SC2034 (warning): FOO appears unused. Verify use (or export if used externally).


In foo.sh line 5:
echo $foo
     ^--^ SC2154 (warning): foo is referenced but not assigned.
     ^--^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
echo "$foo"
...
```

klutchell: Actionlint does not process all files in the repository, only workflow files in `.github/workflows/*.yml`.

Were you expecting `npm run lint` to process `foo.sh`?

ab77: I am expecting it to pick up unquoted env vars in `.github/workflows/flowzone.yml` and from what I can see, it passed them currently.

klutchell: Can you link to these unquoted vars in your PR so I can see what was skipped? What if you introduce a different shellcheck error?

ab77: Even invalid shell passes here, it might be a local workstation issue - can you confirm it flags an unquoted env var on your end?

klutchell: Can confirm it works on my laptop.

```yml
      - name: Toggle auto-merge
        if: steps.is_draft_pr.outputs.result == 'false'
        env:
          <<: *gitHubCliEnvironment
          # DO NOT use the automatic github token (GITHUB_TOKEN) as it will not trigger merge events!
          GH_TOKEN: "${{ steps.gh_app_token.outputs.token || secrets.FLOWZONE_TOKEN }}"
        # As long as we avoid the --admin flag, we should never bypass branch protections
        # See: https://cli.github.com/manual/gh_pr_merge
        run: |
          gh pr merge ${{ github.event.pull_request.number }} --merge --auto || true
          # add a shellcheck lint error
          foo=$(uname -a)
          echo $foo
```
```text
> npm run build && npm run lint                                                   

> flowzone@11.1.9 build
> node build


> flowzone@11.1.9 lint
> docker run --rm -v $(pwd):/repo --workdir /repo rhysd/actionlint:1.6.27 -color

.github/workflows/flowzone.yml:4246:9: shellcheck reported issue in this script: SC2086:info:3:6: Double quote to prevent globbing and word splitting [shellcheck]
     |
4246 |         run: |
     |  
```

Make sure that the lint errors you are introducing are present in `.github/workflows/flowzone.yml` since that's what's being checked. That file is updated with `npm run build` or the pre-commit hooks.

PR #991, opened 2024-04-16
check parent commit for changed files on merge

ab77:
.. attempt to fix [this](https://github.com/balena-io/environment-production/pull/2249#issuecomment-2059281138)

testing:
* https://github.com/balena-io/environment-staging/pull/3339
* https://github.com/balena-io/environment-staging/pull/3340
* https://github.com/balena-io/environment-staging/pull/3341

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1881,6 +1881,9 @@ jobs:
 
       - name: Generate stacks matrix
         id: cloudformation_stacks
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
         run: |
           stacks="$(echo '${{ inputs.cloudformation_templates }}' \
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1881,6 +1881,10 @@ jobs:
 
       - name: Generate stacks matrix
         id: cloudformation_stacks
+        env:
+          MERGE_SHA: ${{ needs.versioned_source.outputs.sha }}
+          DEFAULT_BRANCH: ${{ github.base_ref }}
```
klutchell, line 1886: nit, but this is not necessarily the default branch so we should avoid naming it so
```suggestion
          BASE_BRANCH: ${{ github.base_ref }}
```

klutchell:
LGTM

flowzone.yml
ab77: Cheers, I am still trying to find the two commits I need to diff for both sync and merge contexts, I think I have it now..

PR #1021, opened 2024-04-29
add missing contents write permissions to create tags

ab77:
change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1262,7 +1262,7 @@ jobs:
           <<: *getGitHubAppTokenWith
           permissions: >-
             {
-              "contents": "read",
+              "contents": "write",
               "metadata": "read",
               "pull_requests": "write"
             }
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1262,7 +1262,7 @@ jobs:
           <<: *getGitHubAppTokenWith
           permissions: >-
             {
-              "contents": "read",
+              "contents": "write",
```
klutchell, line 1265: need `admin:write` to push a tag reference in the default branch since it bypasses branch protections

(I think, we should test)

klutchell: I know if we remove admin from versioned_source, we are unable to update the default branch reference, not 100% whether that applies to creating new tag references?

ab77: @klutchell let's maybe try first with just `contents`, if it's not enough, I'll add `admin` as well in the next PR..

Page-: I think tags are separate since they aren't part of that branch, just pointers to commits (that happen to be part of that branch) but I can't say 100%

klutchell (approved):

PR #1023, opened 2024-04-29
only send Zulip message if a human reviewed the release notes

ab77:


flowzone.yml
```diff
@@ -1366,10 +1369,13 @@ jobs:
           comment="$(gh pr view ${{ github.event.pull_request.number }} --json comments \
             --jq '.[].[] | select(.author.login=="${{ needs.versioned_source.outputs.author }}").body')"
 
-          EOF="$(openssl rand -hex 16)"
-          echo "text<<$EOF" >> $GITHUB_OUTPUT
-          echo "${comment}" >> $GITHUB_OUTPUT
-          echo "$EOF" >> $GITHUB_OUTPUT
+          # send message only if a human processed the stubbed release notes, otherwise no-op
+          if ! [[ "$comment" =~ '[only keep the important and rephrase]' ]]; then
```
fisehara, line 1373: @ab77 
 As I'm lacking context, how is this making sure that a human reviewed the release notes?

klutchell, line 1373:
```suggestion
          if ! [[ "$comment" =~ 'only keep the important and rephrase' ]]; then
```
I think the inclusion of the `[...]` will match on any character in your string, when you only want to match the whole phrase.

I could be wrong, it should probably be tested for both pass and fail conditions.

ab77: well, the assumption is that if the boiler plate text is still there, no one has reviewed

klutchell:
It looks like the generated `.github/workflows/flowzone.yml` is out of sync with the one in the root of the repo. Check that `npm run build` in your workspace is updating the generated, and why the husky pre-commit hooks aren't working for you.

Also we should squash those commits so its easier to find and revert this change in the future if needed, or see more clearly via the git blame that these were part of the same changeset.

klutchell:
LGTM!

PR #1035, opened 2024-05-07
add missing GH token with correct permissions

ab77:
* fixes [this](https://github.com/balena-io/environment-production/actions/runs/8990387898/job/24695669680#step:8:51)

change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -593,14 +593,14 @@
       version: v0.14.0
 
   - &setupPython # https://github.com/actions/setup-python
-    name: Setup python
+    name: Setup Python
     id: setup-python
     uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
     with:
       python-version: "3.9"
 
   - &setupPoetry # https://github.com/abatilo/actions-poetry
-    name: Setup poetry
+    name: Setup Poetry
     if: steps.setup-python.outputs.python-version != ''
     uses: abatilo/actions-poetry@7b6d33e44b4f08d7021a1dee3c044e9c253d6439 # v3.0.0
     with:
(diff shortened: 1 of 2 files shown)


klutchell (approved):

ab77:
.. actually, I am a bit lost as to why [this](https://github.com/product-os/flowzone/actions/runs/8986693028/job/24683489154#step:8:53) works, but [this](https://github.com/balena-io/environment-production/actions/runs/8990387898/job/24695669680#step:8:50) doesn't..

as fas as I can tell, they are both running the same code..

flowzone.yml
```diff
@@ -1367,6 +1367,7 @@ jobs:
           github.event.pull_request.merged == true
           && steps.create_tag.outputs.sha != ''
         env:
+          GH_TOKEN: ${{ steps.gh_app_token.outputs.token || secrets.FLOWZONE_TOKEN }}
```
klutchell, line 1370: [This](https://github.com/balena-io/environment-production/blob/fb45587920dd21973295af67b1ba523f525fd1d3/.github/workflows/flowzone.yml#L11-L24) is why

klutchell (changes requested):

flowzone.yml
ab77: yep, I get it now, the default token permissions are used otherwise, which in that workflow are all `read`

klutchell: So we can close this PR right, and just adjust the permissions of the automatic GITHUB_TOKEN on that repo?

I would prefer we use the automatic token where it makes sense to do so, and only use the Flowzone token for cases where special permissions are required (like cross-repo permissions)

ab77:
I'll adjust the token in the repo.

PR #1046, opened 2024-05-13
Use PR body to build GitHub/Balena release notes

pipex:
Look for a `## Release notes` section on the PR body and append the contents of the section to the release notes in GitHub and balenaCloud.

## Release Notes 

Flowzone will now look for a `## Release notes` section on the PR body. When defined, it will use the contents of that section and the PR title for the automatically [generated release notes in GitHub](https://github.com/product-os/flowzone/releases/tag/v11.4.10). If the repository is identified as a [Balena project](https://github.com/product-os/flowzone?tab=readme-ov-file#balena), it will also use the user provided release notes to populate the Balena release notes.

If no `Release notes` section is defined, the workflow will just use the Git change-log, unless the `release_notes` option is set to `false`.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1267,6 +1267,9 @@ jobs:
 
     <<: *rootWorkingDirectory
 
+    outputs:
+      body: ${{ steps.format_release_notes.outputs.body }}
+
     env:
       <<: *gitHubCliEnvironment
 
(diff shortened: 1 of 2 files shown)


pipex:
Here is an example of how this looks like https://github.com/product-os/flowzone/releases/tag/untagged-d51fe7e9266683be6551

flowzone.yml
```diff
@@ -3135,7 +3173,9 @@ jobs:
 
       - *deleteDraftGitHubRelease
       - *checkoutVersionedSha
-      - *getReleaseNotes
+      - <<: *getReleaseNotes
+        env:
+          user_defined: ${{ needs.release_notes.outputs.body }}
```
klutchell, line 3178: I know you added `look_back="${look_back:-1}"` but I think providing both required envs here would be cleaner, and we don't have to look at the function to figure out the default value.

pipex: Done

klutchell:
I think some of those commits could be squashed, but then it looks good

pipex:
Squashed

klutchell (approved):

PR #1048, opened 2024-05-14
Fix renovate release notes

pipex:
The renovate release notes were being overwritten by the default non-renovate behavior.

This updates the code to handle both cases exclusively.


pipex:
I self-certify!

PR #1049, opened 2024-05-14
Reuse notable changes format on release notes

pipex:
The 'Notable changes' section was being formatted differently on the GitHub release notes and the Zulip message.


pipex:
I self-certify!

PR #1050, opened 2024-05-15
Improve notable changes generation

pipex:
A bug in the previous version was ignoring the contents inside the summary tag. This also adds indentation to second level changes (starting with a `> -`) and removes duplicate lines.


pipex:
I self-certify!

PR #1051, opened 2024-05-16
Restore re-itemization of notable changes

pipex:
For the release notes comment, this restores the re-itemization of the notable changes list with `*`, to make it clear that the lines should be edited.

Change-type: patch


pipex:
I self-certify!

PR #1052, opened 2024-05-16
Base versioned commit on the tip of the PR head branch

klutchell:
Historically we have used the PR merge commit, which is the GitHub default, to ensure that we are testing the same merges that will be committed to the target branch.

However, we already require that branches be rebased on the target before merging via branch protections so there is no benefit gained from using the merge commit.

We also would like to enable use cases where we are testing against old versions of the target branch and not include changes from the tip of the target.

Change-type: major
See: https://balena.zulipchat.com/#narrow/stream/348930-balena-io.2Fflowzone/topic/Flowzone.20runs.20tests.20w.2F.20merge.20w.2F.20master/near/438995412
See: https://balena.fibery.io/Work/Project/Base-Flowzone-versioned-commit-on-the-tip-of-the-PR-head-537

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -411,12 +411,12 @@ jobs:
               "metadata": "read",
               "pull_requests": "read"
             }
-      - name: Checkout refs/pull/${{ github.event.number }}/merge
+      - name: Checkout pull request head sha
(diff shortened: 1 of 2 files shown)


klutchell:
Test with a fork of a fork here: https://github.com/balena-io-experimental/flowzone/pull/8

dfunckt (approved):
Nice!

PR #1055, opened 2024-05-16
Restore MD quote for second level changes

pipex:
The extra indentation for the second level changes is lost in the release notes comment, so this restores the previous formatting only for the comment.

The general release notes will look the same.

Change-type: patch


pipex:
I self-certify!

PR #1056, opened 2024-05-18
Allow to install additional cargo packages

pipex:
Adds the `cargo_install` input to allow users to install additional cargo packages before the build.

Change-type: minor


pipex:
Does not work, the installed packages are not available inside the cross environment.

PR #1057, opened 2024-05-19
Fix regression on the GitHub finalize job

pipex:
The work on PR #1046 made the finalize github release job dependent on the `release_notes` input. That's a regression on the correct behavior. This PR moves the PR comment and deploy message actions to a separate job that is dependent on the release notes input, and makes the release_notes job run on every PR, which allows the github finalize job to always run on merge.

This is a major change because it changes the way the release_notes input work, even though it restores the right behavior.

Change-type: major

## Release notes

This release fixes a regression that prevented finalizing the GitHub release unless the `release_comment` input was set to true. This restores the correct behavior, however it is a major change as it is changing an interface (even if the interface was bad).

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1289,13 +1289,13 @@ jobs:
       (
         github.event.action != 'closed' ||
         github.event.pull_request.merged == true
-      ) &&
-      inputs.release_notes == true
+      )
 
     <<: *rootWorkingDirectory
 
     outputs:
       body: ${{ steps.format_release_notes.outputs.body }}
+      comment: ${{ steps.format_release_notes.outputs.comment }}
 
     env:
       <<: *gitHubCliEnvironment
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1289,13 +1289,13 @@ jobs:
       (
         github.event.action != 'closed' ||
         github.event.pull_request.merged == true
-      ) &&
-      inputs.release_notes == true
+      )
```
klutchell, line 1292: I would prefer we keep this job as opt-in, especially with how many changes it's been getting recently.

ab77:
@pipex I think it's fine by looking at the changes; the ultimate test would be to run it against https://github.com/balena-io/environment-production or https://github.com/product-os/environment-production

klutchell (approved):

pipex:
@klutchell just rebased, I think we can merge and test the comments with the production PRs

PR #1080, opened 2024-06-12
minor: generate sbom for node/npm, python/poetry and rust/cargo projects

aethernet:
This should automatically generate an sbom for nodejs projects using cyclonedx and add upload it to the assets.
Once this works, the next step will be to publish it to a central dependency-track server.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -688,6 +688,24 @@
     run: |
       aws eks update-kubeconfig --name "$(echo "${KUBE_CTX}" | awk -F'/' '{print $2}')"
 
+  - &publishSBOMArtifacts
+    name: Publish SBOM artifacts
+    uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+
+  - &publishSBOMToDependencyTrack
+    name: Publish SBOM To Dependency Track
+    if: ${{ env.SERVER_HOSTNAME != '' }}
+    run: |
+      curl -X "PUT" "https://${{ env.SERVER_HOSTNAME }}/api/v1/bom" \
+        -H 'Content-Type: application/json' \
+        -H 'X-API-Key: ${{env.API_KEY}}' \
+         -d '{
+          "projectName": "'"${{env.PROJECT_NAME}}"'",
+          "projectVersion": "'"${{env.PROJECT_VERSION}}"'",
+          "autoCreate": "true",
+          "bom": "'"$(base64 -w 0 "${{ env.BOM_FILE }}")"'"
+        }'
+
   - &convenienceFunctions
     name: Convenience functions
     id: functions
@@ -790,6 +808,9 @@ on:
         required: false
       CUSTOM_JOB_SECRET_3:
         description: "Optional secret for using with custom jobs"
+        required: false
+      DTRACK_TOKEN:
+        description: "API key for Dependency-Track integration"
         required: false
 
     inputs:
@@ -1551,6 +1572,7 @@ jobs:
       has_npm_lockfile: ${{ steps.npm_lock.outputs.has_npm_lockfile }}
       npm_private: ${{ steps.npm.outputs.private }} # can be null or unset
       npm_docs: ${{ steps.npm.outputs.docs }} # can be null or unset
+      npm_sbom: ${{ steps.npm.outputs.sbom }} # can be null or unset
       node_versions: ${{ steps.node_versions.outputs.json }}
       npm_access: ${{ steps.access.outputs.access }}
 
@@ -1567,6 +1589,7 @@ jobs:
             echo "enabled=true" >> "${GITHUB_OUTPUT}"
             echo "private=$(jq -r '.private' package.json)" >> "${GITHUB_OUTPUT}"
             echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> "${GITHUB_OUTPUT}"
+            echo "sbom=true" >> "${GITHUB_OUTPUT}" # TODO: evaluate the need to offer an option to disable SBOM generation
             echo "NODE_VERSIONS=[]" >> "${GITHUB_ENV}"
           else
             echo "enabled=false" >> "${GITHUB_OUTPUT}"
@@ -1878,6 +1901,7 @@ jobs:
       python_poetry: ${{ steps.python_poetry.outputs.enabled }}
       python_versions: ${{ steps.python_versions.outputs.json }}
       pypi_publish: ${{ steps.python_poetry.outputs.pypi_publish }}
+      python_sbom: true # TODO: offer a way to deactivate SBOM generation
 
     steps:
       - *getGitHubAppToken
@@ -1960,6 +1984,7 @@ jobs:
     outputs:
       cargo_targets: ${{ steps.cargo_targets.outputs.build }}
       cargo: ${{ steps.cargo_yml.outputs.enabled }}
+      cargo_sbom: true # TODO: offer a way to deactivate SBOM generation
 
     steps:
       - *getGitHubAppToken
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -2367,6 +2392,42 @@ jobs:
           path: ${{ runner.temp }}/docs.tar.zst
           retention-days: 90
 
+  npm_sbom:
```
klutchell, line 2395: This should be added as a dependency of `all_jobs`

flowzone.yml
```diff
@@ -3016,6 +3099,51 @@ jobs:
         run: |
           poetry run pytest tests/
 
+  python_sbom:
```
klutchell, line 3102: This should be added as a dependency of `all_jobs`

flowzone.yml
```diff
@@ -3365,6 +3493,50 @@ jobs:
           echo "sha_tag=${sha_tag}" >> "${GITHUB_OUTPUT}"
           echo "version_tag=${version_tag}" >> "${GITHUB_OUTPUT}"
 
+  cargo_sbom:
```
klutchell, line 3496: This should be added as a dependency of `all_jobs`

klutchell:
The structure looks good to me in general. What made you decide to go with separate jobs for the sbom generation?

In some cases it makes sense to keep the steps separate, but in this case their seems to be enough duplication between the test jobs and the sbom jobs that we could simply add them as conditional steps at the end of the test job?

It would be faster if they just tacked on to the end of test jobs, and reduce the glut of jobs in the list, but if there are good reasons for keeping them as separate jobs that's fine too.

flowzone.yml
```diff
@@ -2367,6 +2392,42 @@ jobs:
...
           retention-days: 90
 
+  npm_sbom:
+    name: Generate SBOM for NPM
+    runs-on: ${{ fromJSON(inputs.runs_on) }}
+    needs:
+      - is_npm
+      - npm_test
+      - versioned_source
+    if: ${{ needs.is_npm.outputs.npm == 'true' && needs.is_npm.outputs.npm_sbom == 'true' }}
```
klutchell, line 2402: I wish jobs could build the chain of needed dependencies, so you wouldn't need to explicitly mark `is_npm` as needed, but alas it's required in order to call `needs.is_npm`. LGTM.

flowzone.yml
```diff
@@ -686,6 +686,24 @@
     run: |
       aws eks update-kubeconfig --name "$(echo "${KUBE_CTX}" | awk -F'/' '{print $2}')"
 
+  - &publishSBOMArtifacts
+    name: Publish SBOM artifacts
+    uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+
+  - &publishSBOMToDependencyTrack
+    name: Publish SBOM To Dependency Track
+    if: ${{ vars.DTRACK_API != '' }}
```
ab77, line 695: use  `${{ env.SERVER_HOASTNAME }}` everywhere or drop it and just use `${{ vars.DTRACK_API || 'default.hostname.org' }}`

aethernet: good catch

aethernet:
> What made you decide to go with separate jobs for the sbom generation?

Mainly because the test job uses a matrix and there's no need to generate a bom for each version of node or python, etc. Maybe there will be cases where it makes sense, but for now I think we're good with one per project.

I started with a single sbom generator using `if:` to choose the correct language, but it felt cleaner this way.

flowzone.yml
```diff
@@ -2997,6 +3065,21 @@ jobs:
...
+          version=$(grep '^version =' pyproject.toml | sed 's/version = "\(.*\)"/\1/')
+          branch_tag="$(echo "build-${GITHUB_HEAD_REF}" | sed 's/[^[:alnum:]]/-/g')"
+          sha_tag="${branch_tag}-${{ github.event.pull_request.head.sha }}"
+          version_tag="${version}-${branch_tag}-${{ github.event.pull_request.head.sha }}"
+
+          echo "package=${package}" >> "${GITHUB_OUTPUT}"
+          echo "version=${version}" >> "${GITHUB_OUTPUT}"
+          echo "branch_tag=${branch_tag}" >> "${GITHUB_OUTPUT}"
+          echo "sha_tag=${sha_tag}" >> "${GITHUB_OUTPUT}"
+          echo "version_tag=${version_tag}" >> "${GITHUB_OUTPUT}"
```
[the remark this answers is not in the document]
aethernet: Do you mean between languages?

It could probably be done. AFAIU would need to split `package` and `version` in a language-specific step and pass the `version` as an env to the shared one.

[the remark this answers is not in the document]
klutchell: I thought I saw another python job using the same function but I must have been confused. It's fine as it is.

flowzone.yml
```diff
@@ -1567,6 +1589,7 @@ jobs:
             echo "enabled=true" >> "${GITHUB_OUTPUT}"
             echo "private=$(jq -r '.private' package.json)" >> "${GITHUB_OUTPUT}"
             echo "docs=$(jq -r '.scripts | has("doc")' package.json)" >> "${GITHUB_OUTPUT}"
+            echo "sbom=true" >> "${GITHUB_OUTPUT}" # TODO: evaluate the need to offer an option to disable SBOM generation
```
klutchell, line 1592: What are the chances of this process breaking on untested repos? Do we need the option to turn it off, or should it be opt-in at first.

On the other hand, we have very few python and cargo projects so this seems low risk.

aethernet: I'd say low but non-null

I tested python on a copy of balena-cli (works fine), but didn't test the cargo one outside this repo.

aethernet: Here's the result for the latest test atm: https://github.com/product-os/flowzone/actions/runs/9619401059

aethernet:
I think this version is ready (unless you'd prefer for me to add variables to opt-in/out).
Setting up a d-track server and configuring it is not required to start producing these assets.

flowzone.yml
```diff
@@ -2369,6 +2394,41 @@ jobs:
...
+
+      - run: npm install
+
+      - name: Generate SBOM
+        run: |
+          npx @cyclonedx/cyclonedx-npm --output-file=${{ runner.temp }}/bom.xml
+
+      - <<: *publishSBOMArtifacts
+        with:
+          name: sbom-npm-${{ github.event.pull_request.head.sha }}
```
klutchell, line 2420: I don't think there is a need to append the head sha to the artifact name here, it would be much cleaner without it?

flowzone.yml
```diff
@@ -3018,6 +3100,50 @@ jobs:
...
+      - name: Install CycloneDX for Python
+        run: |
+          pip3 install 'cyclonedx-bom>=1.4.0,<4'
+
+      - name: Generate SBOM
+        run: cyclonedx-py -r -i ./poetry.lock --format xml -o ${{ runner.temp }}/bom.xml
+
+      - <<: *publishSBOMArtifacts
+        with:
+          name: sbom-python-${{ github.event.pull_request.head.sha }}
```
klutchell, line 3135: Same here, we can probably drop the sha from the variable name.

flowzone.yml
```diff
@@ -3367,6 +3493,49 @@ jobs:
...
+        run: cargo install cargo-cyclonedx
+
+      - name: Generate SBOM
+        run: |
+          cargo cyclonedx --override-filename bom
+          mv bom.xml ${{ runner.temp }}/bom.xml
+
+      - <<: *publishSBOMArtifacts
+        with:
+          name: sbom-cargo-${{ github.event.pull_request.head.sha }}
```
klutchell, line 3527: Same here, we can probably drop the sha from the variable name but adding a `gh-release-` prefix might be nice to attach to the GH release.

flowzone.yml
klutchell: You would also want `github_publish` to depend on these jobs if you go that route.

aethernet: good call

flowzone.yml
```diff
@@ -2369,6 +2394,41 @@ jobs:
...
+      - run: npm install
+
+      - name: Generate SBOM
+        run: |
+          npx @cyclonedx/cyclonedx-npm --output-file=${{ runner.temp }}/bom.xml
+
+      - <<: *publishSBOMArtifacts
+        with:
+          name: gh-release-sbom-npm
+          path: ${{ runner.temp }}/bom.xml
```
klutchell, line 2421: I believe the artifacts inside the zip need to have unique names or they will overwrite each other when being attached to the release. See the output [here](https://github.com/product-os/flowzone/releases/tag/untagged-16e8c096dae978954854).

aethernet: makes sense

aethernet: It's fixed :) https://github.com/product-os/flowzone/releases/tag/untagged-e866e755aabd0a215bb2

klutchell (approved):
Nice work! 🎉

PR #1090, opened 2024-06-21
Fix string process for dockerhub images

klutchell:
This should re-enable publishing of READMEs
as DockerHub descriptions on finalize.

Change-type: patch


ab77 (approved):
Looks OK, but the ultimate test is whether it works :)

PR #1095, opened 2024-07-02
BREAKING: Enforce workflow shellcheck via actionlint

klutchell:
Change-type: major

This will require someone during the cooldown to help to fix projects with bad shell syntax in their workflows.


klutchell:
@ab77 how do you feel about spending the cooldown period next week fixing (or helping others to fix) workflows with bad syntax?

ab77 (approved):

PR #1097, opened 2024-07-03
patch: add an option to not generate sbom

aethernet:
SBOM generation is failing for some repository with some specific configuration.

This PR adds a global flowzone parameter to turn off sbom generation (on by default) and will `continue-on-error`.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -277,6 +277,11 @@ on:
         type: number
         required: false
         default: 20
+      generate_sbom:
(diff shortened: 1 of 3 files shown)


klutchell:
Should we also add `continue-on-error: true` to the steps that can fail under some conditions?

That way sbom generation is best effort for now, and doesn't require being manually disabled for some repos. The error will still get raised in the job summary but won't fail the job.

klutchell (approved):

aethernet:
good call

I've added it, but will keep the option to disable as there might be repositories where we don't want any sbom created.

PR #1110, opened 2024-07-11
Update setupNode to node v20

joshbwlng:
Change-type: patch


joshbwlng:
Moving to https://github.com/product-os/flowzone/pull/1111

PR #1121, opened 2024-07-18
Fail node install dependencies step if npm install fails

Page-:
Change-type: patch


.github/workflows/flowzone.yml
```diff
@@ -1902,6 +1902,7 @@ jobs:
           npm run flowzone-preinstall --if-present
       - name: Install dependencies
         run: |
+          set -e
```
klutchell, line 1905: The default shell in this job (and most jobs) is already `bash --noprofile --norc -eo pipefail -x {0}` so I think this is noop

PR #1122, opened 2024-07-18
Pin Node 22 minor version to 22.4

pipex:
There is a bug in Node 22.5.0 that is affecting NPM workflows. This is a temporary fix until that issue closes.

Relates-to: npm/cli#7657
Change-type: patch


klutchell:
LGTM

PR #1132, opened 2024-07-22
Checkout versioned source with full depth for release notes

klutchell:
Default to a fetch depth of 1 unless overridden via the input. Always use fetch depth of 0 when release notes are being calculated.

Some balenaOS repositories have submodules that require a fetch depth of 1700+ in order to get the pinned commit. Those repos need to be able to override the default of 1.

Change-type: minor

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -91,8 +91,6 @@
       # prevent git from existing with 141
       set +o pipefail
 
-      # We are limiting search depth but we need tags to calculate the changelog
-      git fetch -q --tags
       look_back="${look_back:-1}"
       previous_tag="$(git --no-pager tag --list --sort=-version:refname "v*.*.*" --merged | head -n${look_back} | tail -n1)"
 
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -218,7 +218,20 @@
...
+      fetch-depth: ${{ inputs.checkout_fetch_depth }}
+      # Note that fetch-tags is not currently working as described:
+      # https://github.com/actions/checkout/issues/1781
+      fetch-tags: true
+      submodules: "recursive"
+      # fallback to an invalid ref if the checkout ref is undefined
+      ref: "${{ needs.versioned_source.outputs.sha || '¯\_(ツ)_/¯' }}"
+      token: ${{ steps.gh_app_token.outputs.token || secrets.FLOWZONE_TOKEN }}
+
+  - &checkoutVersionedShaDeep # https://github.com/actions/checkout
```
pipex, line 230: If we are adding this and using it where we know it's required, why do we still need the input?

klutchell: Because some balenaOS device repos require a full fetch depth for every job with submodules, not just the release note jobs.

So we expose the input, make the default 1, except where we know we need it to be 0 and ignore the input.

klutchell: Repos needing full fetch depth:
- https://github.com/balena-os/balena-trex/actions/runs/10019894679/job/27749653700?pr=184
- https://github.com/balena-os/balena-tgw/commit/56e0bceeee6dd7a0582bd29a8b4e9fab6842e059

pipex (approved):

PR #1135, opened 2024-07-24
Add support for windows and darwin npm builds and tests

klutchell:
Rather than manually testing all engine combinations ahead
of time we can use a hardcoded JSON matrix for npm tests.

Now we can support multiple platforms and node versions combinations
and simply skip the tests if unsupported by platform.json.

The publish job will automatically choose the artifact generated
by the newest node version based on the matrix order.

The npm SBOM will be published for each valid combination.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -640,15 +640,6 @@
     with:
       version: "2.15.43"
 
-  - &sortNodeVersions
-    name: Sort node versions
-    id: node_versions
-    env:
-      VERSIONS: ${{ needs.is_npm.outputs.node_versions }}
-    run: |
-      echo "min=$(echo "${VERSIONS}" | jq -r '.[]' | sort --version-sort | head -n1)" >> "${GITHUB_OUTPUT}"
-      echo "max=$(echo "${VERSIONS}" | jq -r '.[]' | sort --version-sort --reverse | head -n1)" >> "${GITHUB_OUTPUT}"
-
   - &waitForCloudFormation
     name: Wait for resources
     run: |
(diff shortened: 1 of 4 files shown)


flowzone.yml
```diff
@@ -2365,114 +2317,93 @@ jobs:
...
+        if: needs.is_npm.outputs.npm_docs == 'true' && steps.check_engine.outcome == 'success'
         run: tar --auto-compress -cvf ${{ runner.temp }}/docs.tar.zst ./docs
 
       - name: Upload artifact
-        if: needs.is_npm.outputs.npm_docs == 'true' && steps.node_versions.outputs.max == matrix.node_version
+        if: needs.is_npm.outputs.npm_docs == 'true' && steps.check_engine.outcome == 'success'
         uses: actions/upload-artifact@0b2256b8c012f0828dc542b3febcab082c67f72b # v4.3.4
         with:
-          name: docs-${{ github.event.pull_request.head.sha }}-${{ matrix.node_version }}
+          name: docs-${{ github.event.pull_request.head.sha }}-${{ strategy.job-index }}
```
pipex, line 2381: The job-index is relative to the matrix right? Are the jobs always ran in the same order? What could cause it to change? I'm thinking of possible ways an unmerged PR could fail on merge because of a flowzone change between the two events

flowzone.yml
```diff
@@ -2365,114 +2317,93 @@ jobs:
...
         run: npm run doc
 
       - name: Compress docs
-        if: needs.is_npm.outputs.npm_docs == 'true' && steps.node_versions.outputs.max == matrix.node_version
+        if: needs.is_npm.outputs.npm_docs == 'true' && steps.check_engine.outcome == 'success'
         run: tar --auto-compress -cvf ${{ runner.temp }}/docs.tar.zst ./docs
 
       - name: Upload artifact
-        if: needs.is_npm.outputs.npm_docs == 'true' && steps.node_versions.outputs.max == matrix.node_version
+        if: needs.is_npm.outputs.npm_docs == 'true' && steps.check_engine.outcome == 'success'
```
pipex, line 2378: I wonder if we could skip some of these steps since only one artifact will be used in the end

pipex:
This looks like a great improvement, just added a couple of questions

flowzone.yml
klutchell: I confirmed the order is always the same and based on the order in which the values are provided in the matrix arrays.

flowzone.yml
klutchell: I wanted to, but as all the jobs are running in parallel they have no way of knowing whether their artifact is the one that will be used or not.

This is the downside of performing the engine checks as part of the tests, we don't know ahead of time which matrix jobs are expected to pass and create artifacts. I still think the pros outweigh the cons though, especially since these uploads are done in parallel.

PR #1162, opened 2024-08-15
CloudFormation package does not install Lambda deps.

ab77:
Tested [here](https://github.com/balena-io/environment-staging/actions/runs/10409345156/job/28831503909#step:16:52)

There needs to be an additional step when packaging CFN tempates containing Lambda/Code: reference directory. By default, without using a completely different CFN "SAML" workflow (we don't want to go there maybe yet), dependencies need to be installed into the same directory as the Lamda code.

This comit implements this workflow for Python code when requirements.txt is present.

change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -3318,7 +3318,7 @@ jobs:
           accountId: ${{ secrets.CF_ACCOUNT_ID }}
           wranglerVersion: "3.5.1" # latest - https://www.npmjs.com/package/wrangler
           command: pages deploy --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} build/
-      
+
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -4086,6 +4086,26 @@ jobs:
           with_backoff aws cloudformation validate-template \
             --template-url "https://s3.amazonaws.com/${{ steps.make_bucket.outputs.s3_bucket }}/${tmpvalid}"
 
+      - name: Prepare (Python) Lambda dependencies
+        run: |
+          source '${{ steps.functions.outputs.with_backoff }}'
```
klutchell, line 4091: Is this line needed?

ab77: Good catch, nothing to retry in this step

klutchell (approved):

PR #1169, opened 2024-08-22
Check if Code does in fact point to a valid directory

ab77:
`AWS::Lambda::Function(s)` can have different code syntax

(1) point to a directory to be packaged manually

```
  LambdaFunction:
    Type: AWS::Lambda::Function
    Properties:
      Code: lambda/
      Handler: handler.lambda_handler
```

(2) inline

```
  BackupCopyManagerLambda:
    Type: AWS::Lambda::Function
    Properties:
      Handler: index.lambda_handler
      Code:
        ZipFile: |
          import time, json, boto3, os, sys, logging, urllib3

          COPY_TO_ACCOUNT_TAG = os.environ.get('COPY_TO_ACCOUNT_TAG')
          ...
```

This patch should handle both cases (currently it only handles 1).

change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -4090,18 +4090,20 @@ jobs:
         run: |
           template_base_directory="$(dirname '${{ steps.shared.outputs.template_file }}')"
 
-          python_lambdas="$(cat '${{ steps.shared.outputs.template_file }}' \
+          python_lambdas="$(cat <'${{ steps.shared.outputs.template_file }}' \
             | yq e -oj \
             | jq -r '.Resources[]
             | select((.Type=="AWS::Lambda::Function")
             and (.Properties.Runtime | startswith("python"))).Properties.Code')"
 
           for python_lambda in ${python_lambdas}; do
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -4090,18 +4090,20 @@ jobs:
         run: |
           template_base_directory="$(dirname '${{ steps.shared.outputs.template_file }}')"
 
-          python_lambdas="$(cat '${{ steps.shared.outputs.template_file }}' \
+          python_lambdas="$(cat <'${{ steps.shared.outputs.template_file }}' \
```
klutchell, line 4093: What is this change helping with?

ab77: There's some fascinating reading [here](https://unix.stackexchange.com/questions/258931/difference-between-cat-and-cat), especially in the comments and pointing to more SO posts.

I am standardising on `cat <input` and `cat >output`, so whenever I see a plain `cat` I change it.

klutchell (approved):

PR #1181, opened 2024-09-12
Remove node 16 as it went EOL on 2023-09-11

Page-:
Change-type: patch


Page-:
Just to cover it preemptively, yes the 2023 is correct which means it's been EOL for just over a year now

PR #1221, opened 2024-10-17
BREAKING: Use a single Node.js version for test & publish

klutchell:
Projects should be opinionated and define the desired node version in .node_version in the root of the project. This is supported natively via the setup-node action to and renovate to auto-detect the version to use.

Failing this, we will still fall back to testing with recent LTS releases but only select the first compatible version detected, in decending order (newest first).

Change-type: major
See: https://balena.fibery.io/Work/Improvement/Test-Node.js-projects-against-a-single-version-2262


klutchell:
Discussed with the team and the gains by dropping some Node.js test variants would be negligible while we would lose valuable test coverage.

PR #1251, opened 2024-11-11
Website publish requires issues:write

ab77:
Fixes: https://github.com/balena-io/balena-cloud/actions/runs/11783098446/job/32825587737#step:12:14

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -3367,6 +3367,7 @@ jobs:
             {
               "metadata": "read",
               "contents": "read",
+              "issues": "read",
               "pull_requests": "write"
             }
       - *checkoutVersionedSha
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -3509,6 +3509,7 @@ jobs:
             {
               "metadata": "read",
               "contents": "read",
+              "issues": "read",
               "pull_requests": "write"
             }
       - name: Checkout versioned commit


flowzone.yml
```diff
@@ -3367,6 +3367,7 @@ jobs:
             {
               "metadata": "read",
               "contents": "read",
+              "issues": "write",
```
klutchell, line 3370: `"pull_requests": "write"` should be enough, based on testing I did elsewhere.

klutchell: Is this not enough? https://github.com/balena-io/balena-cloud/blob/e0a5401062a8e8799227e9d8f231fbf1d3078366/.github/workflows/flowzone.yml#L21C1-L21C49

klutchell: Hmm, it is using the [listComments](https://github.com/peter-evans/find-comment/blob/main/src/find.ts) REST endpoint which may need more than actually posting a comment. @ab77 why don't we try just giving it "issues": "read"?

ab77: I just took the settings from [here](https://github.com/peter-evans/create-or-update-comment)

image

klutchell: I think try it with just `issues:read` and that is hopefully enough.

klutchell: Let's try this instead: https://github.com/product-os/flowzone/pull/1252

PR #1254, opened 2024-11-12
Use app installation token in event_types and docker_publish jobs

rcooke-warwick:
These requested more than a default: none permissions from the workflow GITHUB_TOKEN - which fails if the caller workflow doesn't at least these permissions

Change-type: patch

Unsure if this will fix - also if there are other places where this will need to be done


klutchell:
Superseded by https://github.com/product-os/flowzone/pull/1255

PR #1255, opened 2024-11-12
Use App Installation token where read-only is not sufficient

klutchell:


flowzone.yml
```diff
@@ -3362,7 +3371,9 @@ jobs:
           permissions: >-
             {
               "metadata": "read",
-              "contents": "read"
+              "contents": "read",
+              "issues": "read",
+              "pull_requests": "write"
```
Page-, line 3376: It might be worth adding comments about why particular write permissions are needed

flowzone.yml
```diff
@@ -1064,13 +1065,21 @@ jobs:
...
       - *rejectInternalPullRequestTarget
+      - *rejectMissingSecrets
+      - <<: *getGitHubAppToken
+        with:
+          <<: *getGitHubAppTokenWith
+          permissions: >-
+            {
+              "metadata": "read",
+              "issues": "read",
+              "pull_requests": "write"
```
Page-, line 1081: It might be worth adding comments about why particular write permissions are needed

Page- (approved):

PR #1266, opened 2024-11-18
Improve enforcement of linear commit history

klutchell:
Use the GitHub API to check commit parents in BASE and HEAD
branches and ensure we are following a linear history.

HEAD should not have any merge commits, so reject any HEAD branches
where one or more commits have multiple parents.

The PR merge commit must include the HEAD commit as a parent,
otherwise reject the versioning job.

Change-type: minor

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -571,6 +571,56 @@
         Please consult the documentation for more information."
       exit 1
 
+  - &rejectNonLinearHead
+    # This can happen when a PR is updated with a merge commit from main, rather than rebased on main.
+    # In the GitHub UI this is represented by the "Update with merge commit" option vs "Update with rebase".
+    # https://github.com/actions/github-script
+    # https://octokit.github.io/rest.js/v21/#pulls-list-commits
+    name: Reject HEAD branches containing merge commits
+    if: github.event.pull_request.state == 'open'
+    uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+    with:
+      github-token: ${{ github.token }}
+      result-encoding: json
+      script: |
+        const { data: commits } = await github.rest.pulls.listCommits({
+          ...context.repo,
+          pull_number: context.payload.pull_request.number
+        });
+
+        if (commits.some(({ parents }) => parents.length > 1)) {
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1137,6 +1189,13 @@ jobs:
       <<: *gitHubCliEnvironment
 
     steps:
+
+      - *getHeadBranchMergeCommits
+      - *rejectNonLinearHead
```
Page-, line 1194: These seem like they must be run together and in order so I'm wondering why they aren't a single anchor since they are so 1-1 dependent?

flowzone.yml
```diff
@@ -571,6 +571,56 @@
...
+        console.debug(`HEAD commit SHA: ${context.payload.pull_request.head.sha}`)
+
+        const { data: commit } = await github.rest.repos.getCommit({
+          ...context.repo,
+          ref: context.sha
+        });
+
+        console.debug('Commit parents: %s', JSON.stringify(commit.parents, null, 2))
+
+        if (!commit.parents.map(parent => parent.sha).includes(context.payload.pull_request.head.sha)) {
```
Page-, line 620: This is slightly more efficient as:
```suggestion
        if (!commit.parents.some(parent => parent.sha === context.payload.pull_request.head.sha)) {
```
And more importantly imo is also a little clearer as to the intent. I'd also probably use destructuring to format it as
```suggestion
        if (!commit.parents.some(({ sha }) => sha === context.payload.pull_request.head.sha)) {
```

flowzone.yml
```diff
@@ -571,6 +571,56 @@
...
+    with:
+      github-token: ${{ github.token }}
+      result-encoding: json
+      script: |
+        const { data: commits } = await github.rest.pulls.listCommits({
+          ...context.repo,
+          pull_number: context.payload.pull_request.number
+        });
+
+        if (commits.some(commit => commit.parents.length > 1)) {
```
Page-, line 591: A minor thing but I'd probably use destructuring here as well like
```suggestion
        if (commits.some(({ parents }) => parents.length > 1)) {
```
so that the important part is a little shorter/less context to reparse (since we already have the context of `commits.some` to show we're iterating those)

PR #1275, opened 2024-11-22
Update qemu to 9.1.2

pipex:
We need the new version to get the fix for https://gitlab.com/qemu-project/qemu/-/issues/1729 which is blocking balena-io-modules/node-systemd#11

Change-type: patch
Relates-to: balena-os/balena-supervisor#2356


pipex:
Tests pass here, but maybe they should not. We need a new qemu version here https://hub.docker.com/r/tonistiigi/binfmt/tags

QEMU setup fails with 

```
Error: Error response from daemon: manifest for tonistiigi/binfmt:qemu-v9.1.2 not found: manifest unknown: manifest unknown
```

See https://github.com/balena-io-modules/node-systemd/actions/runs/11979780007/job/33402751420

PR #1301, opened 2024-12-26
Use github-script to generate versioned blobs and commits

klutchell:
Resolves 10 code-scanning warnings.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -66,62 +66,69 @@
     run: |
       echo "datetime=$(date +"${FORMAT}")" >> $GITHUB_OUTPUT
 
-  # https://docs.github.com/en/rest/git/tags?apiVersion=2022-11-28#create-a-tag-object
+  # https://octokit.github.io/rest.js/v21/#git-create-tag
+  # https://docs.github.com/en/rest/git/tags#create-a-tag-object
   - &createTagObject
     name: Create tag object
-    if: inputs.disable_versioning != true
     id: create_tag
     env:
-      GH_TOKEN: ${{ steps.gh_app_token.outputs.token || secrets.FLOWZONE_TOKEN }}
       TAG: ${{ steps.versionist.outputs.tag }}
       MESSAGE: ${{ steps.versionist.outputs.tag }}
       SHA: ${{ steps.create_commit.outputs.sha }}
-    run: |
-      response="$(gh api -X POST repos/$GH_REPO/git/tags \
-        -F "tag=${TAG}" \
-        -F "message=${MESSAGE}" \
-        -F "object=${SHA}" \
-        -F "type=commit")"
+    uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+    with:
+      github-token: ${{ steps.gh_app_token.outputs.token || secrets.FLOWZONE_TOKEN }}
+      result-encoding: json
+      script: |
+        const { data: tag } = await github.rest.git.createTag({
+          ...context.repo,
+          tag: process.env.TAG,
+          message: process.env.MESSAGE,
+          object: process.env.SHA,
+          type: 'commit'
+        });
+        core.setOutput('sha', tag.sha);
+        return tag;
 
-      echo "$response" | jq .
-      echo "sha=$(echo $response | jq -r .sha)" >> "${GITHUB_OUTPUT}"
-      echo "json=$(echo $response | jq -c .)" >> "${GITHUB_OUTPUT}"
-
-  # https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#update-a-reference
+  # https://octokit.github.io/rest.js/v21/#git-update-ref
+  # https://docs.github.com/en/rest/git/refs#update-a-reference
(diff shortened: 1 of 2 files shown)


klutchell:
Manually tested post-merge steps [here](https://github.com/balena-io-experimental/flowzone/actions/runs/12528311350/job/34943011252). ✅

flowzone.yml
```diff
@@ -1271,78 +1278,96 @@ jobs:
...
+            const { execSync } = require('child_process');
+            const fs = require('fs');
+
+            // Get modified and untracked files
+            const getModifiedFiles = () => {
+              const modifiedCmd = 'git diff --name-only';
+              const untrackedCmd = 'git ls-files --others --exclude-standard';
+              const modified = execSync(modifiedCmd).toString().trim().split('\n');
+              const untracked = execSync(untrackedCmd).toString().trim().split('\n');
+              return [...new Set([...modified, ...untracked])].filter(f => f);
```
pipex, line 1301: What are you filtering here? Null lines? Blank lines?

klutchell: Both, either, I just have trust issues processing git command line output.

Is there a better way?

flowzone.yml
```diff
@@ -1271,78 +1278,96 @@ jobs:
...
+                  encoding: 'base64'
+                });
+
+                entries.push({
+                  path: file,
+                  mode: '100644',
+                  type: 'blob',
+                  sha: blob.sha
+                });
+              }
```
pipex, line 1326: If the size of the files array is reasonable, you might be able to speed this up by using async `readFile` and `Promise.all`, something like

```
const entries = await Promise.all(files.map(async (file) => {
                core.info(`Creating blob for file ${file}...`);
                // Read file content and create blob
                const content = await fs.readFile(file, "utf8");
                const { data: blob } = await github.rest.git.createBlob({
                  ...context.repo,
                  content: Buffer.from(content).toString("base64"),
                  encoding: 'base64'
                });

                return {
                  path: file,
                  mode: '100644',
                  type: 'blob',
                  sha: blob.sha
                };
});
```

If the files array is too large, then you'll get rate limited, but you could use something like [what we do on base images](https://github.com/balena-io-library/base-images/blob/master/scripts/utils.js#L19-L36) to split the calls

flowzone.yml
```diff
@@ -1351,13 +1376,19 @@ jobs:
 
       # update the existing branch(head) name reference(pointer) to the new commit object sha
       - <<: *updateGitReference
+        if: |
+          github.event.pull_request.merged &&
```
pipex, line 1380: Why was this not needed before? Was it a bug?

flowzone.yml
```diff
@@ -1271,78 +1278,96 @@ jobs:
...
+              tree: process.env.TREE_SHA,
+              parents: [process.env.PARENT_COMMIT_SHA]
+            });
+            core.setOutput('author', commit.author.name);
+            core.setOutput('sha', commit.sha);
+            return commit;
 
       # create a new tag object
       - <<: *createTagObject
+        if: inputs.disable_versioning != true
```
pipex, line 1370: Why wasn't this needed before? Was it a bug?

pipex (commented):
Just left some minor comments, nothing that should hold back the merge. 

Don't really know the GitHub API though. Took a quick look at the documentation and what you have seems correct. I assume the PR would not merge otherwise anyway

flowzone.yml
pipex: No, it's fine to filter, I'm just asking because I don't totally know the commands output. This is filtering nullish values, including empty strings, so it is technically correct. 

If you also wanted to filter blank lines (all whiltespace), you would have to do something like `f != null && !!f.trim()`

flowzone.yml
klutchell: It used to be on line 72, as part of the anchor. It makes more sense to have the condition where it is used.

flowzone.yml
klutchell: It used to be on line 93, as part of the anchor. It makes more sense to have the condition where it is used.

flowzone.yml
klutchell: It should never be more than 4-6 files, I'll make the change.

flowzone.yml
klutchell: But the current still filters empty strings, undefined, and null, correct? Just not whitespace?

flowzone.yml
klutchell: I pushed a new commit, if it looks correct and tests pass I'll squash it.

klutchell:
Testing PR open, sync, and merge here: https://github.com/balena-io-experimental/flowzone/pull/12

flowzone.yml
pipex: Yes, that's correct

klutchell:
Post-merge workflow test run: https://github.com/balena-io-experimental/flowzone/actions/runs/12562179536

klutchell:
Merge workflow looks good, everything is versioned and tagged correctly. I've squashed the commits and it's ready to go!

pipex:
LGTM

PR #1303, opened 2024-12-29
Remove unnecessary helper jobs

klutchell:


flowzone.yml
```diff
@@ -3338,7 +3272,7 @@ jobs:
     if: |
       !failure() && !cancelled() &&
       needs.is_balena.result == 'success' &&
-      github.event.pull_request.state == 'open'
+      (github.event.action != 'closed' || github.event.pull_request.merged == true)
```
pipex, line 3275: I assume this means it will run on both PR refresh and merge right?

klutchell: Yes, and push events. So same behaviour as before, and same rules as website_publish

pipex (approved):

PR #1315, opened 2025-01-08
Cleanup is_npm job steps and conditions

klutchell:
Replace multiple shell steps with a single github-script and skip the job entirely if package.json is not found.

Change-type: minor


ab77 (approved):
Looks generally fine to me.

flowzone.yml
```diff
@@ -1909,112 +1887,112 @@ jobs:
...
-          then
-            echo "json=${NODE_VERSIONS}" >> "${GITHUB_OUTPUT}"
-          fi
+        env:
+          NODE_VERSIONS: ${{ toJSON(steps.*.outputs.node_version) }}
+        with:
+          result-encoding: json
+          script: |
+            const { compareVersions } = require('util');
+            const versions = JSON.parse(process.env.NODE_VERSIONS) || '["20.x"]';
```
Page-, line 1990: This will throw if `NODE_VERSIONS === ''`, I don't know if that's a problem/possible though

flowzone.yml
```diff
@@ -1909,112 +1887,112 @@ jobs:
...
+            const sorted = [...versions].sort(compareVersions);
+
+            core.setOutput('min', sorted[0]);
+            core.setOutput('max', sorted[sorted.length - 1]);
```
Page-, line 1994:
```suggestion
            core.setOutput('max', sorted.at(-1));
```

flowzone.yml
klutchell: I was meaning to test that, I'll do so know and handle the throw better. Good catch!

klutchell: Tested [here](https://github.com/product-os/flowzone/actions/runs/12695225997/job/35386913030?pr=1315).

It did not throw, because NODE_VERSIONS was `[]`, but it also failed to fall back to `["20.x"]` so I've pushed a change to re-test.

PR #1317, opened 2025-01-08
Pin helper jobs to ubuntu-24.04 runners by default

klutchell:
This is the new default "ubuntu-latest" tag provided by GitHub starting in Dec 2024.

See: https://github.com/actions/runner-images/issues/10636


pipex:
LGTM

PR #1324, opened 2025-01-16
Use Node major version for matrix and artifact name

pipex:
The node version matrix uses full version (v22.12.0), for the install process and the build artifact name. This can cause an issue if the runners are using different Node versions

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1956,7 +1956,7 @@ jobs:
         name: Check engine
         id: check_engine_18
         run: |
-          node_version="$(node --version)"
+          node_version="$(node --version | cut -d'.' -f1).x"
           if npx -q -y -- check-engine
           then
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1956,7 +1956,7 @@ jobs:
         name: Check engine
         id: check_engine_18
         run: |
-          node_version="$(node --version)"
+          node_version="$(node --version | cut -d'.' -f1).x"
```
klutchell, line 1959: I would rather we use the input that was provided to the Setup node step, instead of always assuming we want to truncate with .x

klutchell (changes requested):

flowzone.yml
pipex: I thought about that, so we would pass the input to `checkEngine` as well

pipex:
Closing in favor of #1325

PR #1325, opened 2025-01-16
Remove version from npm artifact

pipex:
There is only one NPM artifact published during the npm_test job. There is no need for the version suffix.

Change-type: patch


pipex:
Closing in favor of #1326 which achieves the same without breaking existing workflows

PR #1328, opened 2025-01-16
Prevent linting jobs from running on PR close events

klutchell:
#### Prevent linting jobs from running on PR close events

There is no value in running actionlint or pre-commit checks
on PR merge or close events.

Same with the Node version checks, no need on close events.

The octoscan checks should run on merge to update the main
branch code scanning results, but not on PR close.

Tested close event [here](https://github.com/product-os/flowzone/actions/runs/12817468822).
Note that `File List` runs without failing, so we can leverage it for `Custom always` job conditions in a future PR.

#### Use shallow checkout with github token for linting jobs
These jobs do not rely on versioned source, so they
can default to the pull request HEAD sha and fallback
to the github ref for other event types.

Using the github ref is not appropriate for pull_request_target
events where the ref/sha is always the base branch.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -185,6 +185,21 @@
       ref: "${{ needs.versioned_source.outputs.sha || '¯\_(ツ)_/¯' }}"
       <<: *checkoutAuth
 
+  - &shallowCheckout # https://github.com/actions/checkout
+    name: Checkout source
+    uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    with:
+      # Disable submodules and deep cloning.
+      fetch-depth: 1
+      submodules: false
+      # Do not persist credentials.
+      persist-credentials: false
+      # Use the automatic actions token with contents:read permissions
+      token: ${{ github.token }}
+      # Use the tip of the pull request branch for pull request (target) events.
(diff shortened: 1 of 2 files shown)


klutchell:
Confirmed the first commit is correctly skipping jobs on close.
image

flowzone.yml
```diff
@@ -1890,12 +1908,18 @@ jobs:
 
   # check if the repository has a package.json file and which engine versions are supported
   is_npm:
-    name: Check NodeJS versions
+    name: request branch for pull requestJS versions
```
pipex, line 1911:
```suggestion
    name: request branch for pull request JS versions
```

klutchell: That must have been a copy-paste error, I didn't even mean to update that job name? Weird.

pipex: I thought it was weird too, but I assumed you had some reason? :p

pipex (approved):
LGTM

flowzone.yml
klutchell: That's very nice of you

klutchell:
As intended, octoscan and check nodeJS versions still runs on merge, backed by File List.

However linting jobs like pre-commit and actionlint are skipped as planned.

image

PR #1332, opened 2025-01-20
Use github-script to generate release notes

klutchell:
This uses the GitHub API to get previous tags
and commit messages, while avoiding issues due to
extremely long strings and special characters.

This also means we don't need to clone the repo
with the last n tags in order to generate release
notes so we can change the default checkout depth
in a future PR.

This PR also fixes the run conditions for release_notes
to ensure we skip those jobs on repos with versioning disabled.

Example release from this PR: https://github.com/product-os/flowzone/releases/tag/untagged-71d2fb3c1318e8ec115d

## Release notes

Use the GitHub API to get previous tags
and commit messages, while avoiding issues due to
extremely long strings and special characters.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -670,6 +670,8 @@ jobs:
     needs:
       - versioned_source
     if: |
+      inputs.release_notes == true &&
+      needs.versioned_source.outputs.tag != '' &&
       (
         github.event.action != 'closed' ||
         github.event.pull_request.merged == true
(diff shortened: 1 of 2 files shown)


klutchell:
Additional testing here: https://github.com/klutchell/dnscrypt-proxy-docker/pull/520

ab77 (approved):

pipex:
@klutchell the list of commits was not populated https://github.com/product-os/flowzone/releases/tag/v19.7.0

klutchell:
> @klutchell the list of commits was not populated https://github.com/product-os/flowzone/releases/tag/v19.7.0

Good catch, here is the fix https://github.com/product-os/flowzone/pull/1335

PR #1342, opened 2025-01-22
Set correct npm_docs output value

pipex:
The output should have the value `true` if the doc script exists

Change-type: patch


pipex:
Here is a test https://github.com/balena-io/contrato/actions/runs/12918036884/job/36025803642

klutchell (approved):

PR #1351, opened 2025-01-29
Enforce .dockerignore for Docker projects

ab77:
Avoid future instances of leaking credentials in Git history.

change-type: patch


klutchell (changes requested):
We should leverage the existing file_list job for stuff like this.

More importantly, how many of our projects will this break if it merges? Seems like a seriously breaking change and we should print an error or warning but not bail out.

flowzone.yml
```diff
@@ -2091,6 +2091,14 @@ jobs:
       # https://github.com/actions/checkout
       - *shallowCheckout
 
+      # enforce .dockerignore requirement and warn (for now) if .git isn't excluded
+      - name: Check for .dockerignore
+        if: |
+          !contains(needs.file_list.outputs.workdir, '.dockerignore')
+        run: |
+          echo "::error:: Docker builds require .dockerignore file."
+          exit 1
```
klutchell, line 2100: I don't think we should fail the build for this, since good practice would be to just avoid `COPY .` in the Dockerfile, and thus no need for a `.dockerignore` at all.

If anything, I would suggest we look for an existing linter or pre-commit hook that can detect  both `COPY .` AND the lack of a `.dockerignore` file, which is where we have a problem.

Otherwise I think we just train everyone to ignore this error/warning.

klutchell (changes requested):

PR #1358, opened 2025-02-07
Run balena-versionist via github-script

klutchell:
This provides better control over the output and
returns the new version directly.

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1213,38 +1213,39 @@ jobs:
             "$(npm root -g)"/versionist/scripts/generate-changelog.sh .
           fi
 
-      # run balena-versionist
-      # print the error if anything fails
-      # fail the workflow if the error is anything other than "No such file or directory"
-      - name: Run versionist
+      # https://github.com/actions/github-script
+      - name: Run balena-versionist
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1213,38 +1213,39 @@ jobs:
...
+          script: |
+            const path = require('path');
+            const { execSync } = require('child_process');
 
-          versions=()
-          [ -f .versionbot/CHANGELOG.yml ] && read -r -a versions < <(yq e '.[0].version' .versionbot/CHANGELOG.yml)
-          semver="${versions[0]}"
+            // Get global node_modules path
+            const globalNodeModules = execSync('npm root -g').toString().trim();
+            const { runBalenaVersionist } = require(path.join(globalNodeModules, 'balena-versionist'));
```
pipex, line 1232: this seems very convoluted to search for the global module to import it. Isn't there a way to import external modules from github script?

klutchell: Yeah, it seemed weird to me too. Let me see if I can find the recommended approach.

klutchell: I switched to setting NODE_ENV once in the job environment and that seems cleaner.

pipex (approved):
LGTM

PR #1360, opened 2025-02-07
Use hw execution for cp15 barrier instructions

pipex:
ARMv6 builds may hang when building on ARMv8 hardware due to emulated CP15 builds. This configures the runner kernel to use hardware execution rather than emulation to avoid the issue.

Change-type: patch


klutchell (approved):
Nice one!

PR #1366, opened 2025-02-10
Don't assume every project includes repo.yml

ab77:
> e.g. https://github.com/product-os/.github/actions/runs/13248425179/job/36986287233#step:11:41

Check error message string to restore behaviour prior to https://github.com/product-os/flowzone/pull/1358

.. or we add `repo.yml` to **every** project; or refactor `balena-versionist` to have some sensible defaults, error message handling and exit codes

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -512,7 +512,9 @@ jobs:
                 silent: false,
               });
             } catch (e) {
-              core.setFailed(e.message);
+              if (! /^No such file or directory:.*$/.test(e.message)) {
+                core.setFailed(e.message);
+              }
             }
 
             console.log('New version: ', version);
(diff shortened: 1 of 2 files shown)


klutchell (commented):
I don't think trapping the error code will work because the versioning step will have already bailed out.

Either way, I think we need to address this in balena-versionist or inject a repo.yml via Flowzone if both repo.yml and package.json don't exist, as you suggested.

There is a shaping pitch this cycle to pull versionist out of Flowzone into a separate app so that's likely where the above work will happen.

For now, I've been adding repo.yml to the repos that need it.

klutchell:
I've started adding missing repo.yml files where required. It seems like a pretty short list actually, and I will look at fixing the versionist behaviour when we start to extract that behavioiur.

PR #1419, opened 2025-03-18
patch: Always update stacks if stack config is changed

ab77:


flowzone.yml
```diff
@@ -2834,7 +2834,7 @@ jobs:
 
               // Keep stack if template is modified or workflow is modified
               return modifiedFiles.some(file =>
-                file === template || file.startsWith('.github/workflows/')
+                file === template || file.startsWith('.github/workflows/' || file == "aws-cf-templates.yaml")
```
klutchell, line 2837: We also support yml and json extensions

klutchell (approved):

PR #1438, opened 2025-04-01
Use full semver to pin actions/github-script

klutchell:
This allows Renovate to separate the PRs by major/minor/patch revisions instead of just treating them all as digest updates.


pipex:
LGTM

PR #1464, opened 2025-04-24
Remove tag push events from release notes generation

klutchell:
Tag push events are only somewhat supported, and there is no situation where this ever would have worked since we need to look at the PR body and commits to generate notes.

Tag push logs:
image


pipex:
LGTM

PR #1489, opened 2025-05-26
Download versioned source rather than re-clone

klutchell:
This significantly reduces LOC, and improves performance and security
by having to generate far fewer application tokens.

See: https://balena.fibery.io/Work/Improvement/Use-local-artifacts-for-versioned-source-2922

**REVIEWERS**

Expand the diff of `.github/workflows/flowzone.yml` to review the final changes, as some YAML anchors were moved without changing content.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -680,40 +680,41 @@ jobs:
         if: |
           github.event.pull_request.merged &&
           steps.create_tag.outputs.sha
-      - name: Checkout versioned commit
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
-        with:
-          fetch-depth: ${{ steps.git_describe.outputs.depth || 0 }}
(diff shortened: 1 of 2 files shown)


klutchell:
@pipex see anything of concern here? It seems almost too simple.

klutchell:
> @klutchell this looks good to me, I think all jobs already depend on `versioned_source` so that should not cause issues. Will this break custom actions? https://github.com/balena-io/balena-cli/blob/master/.github/actions/publish/action.yml#L30-L35

No, that custom artifact is unrelated to the versioned source artifact we will be publishing here.

Though that specific action you linked should probably be clearing the workspace before unpacking a custom tarball. This PR wouldn't change that though.

pipex (approved):
Nice work! This heavily reduces the number of token requests needed.

klutchell:
Testing merge event here: https://github.com/balena-io-experimental/flowzone/actions/runs/15284827977

PR #1507, opened 2025-06-09
Restore submodule fetch for yocto-based versioning

klutchell:
Submodules are required to get the meta-balena version in yocto-based device repos.

Change-type: patch


kb2ma:
@klutchell  This is just to revert b79b2bb , and there is no impact on the intervening commits?

kb2ma:
LGTM

PR #1509, opened 2025-06-11
Fix release notes generation

pipex:
Release notes were not correctly including all the text in the markdown section. This fixes the regular expression to allow to correctly generate the notes.

Change-type: patch


pipex:
Tested here https://github.com/balena-io-modules/mahler-rs/actions/runs/15593087509/job/43916846024

klutchell (approved):

PR #1574, opened 2025-09-24
Publish cargo workspaces

pipex:
Add support for publishing workspaces using the `--workspace` feature
(made stable as of v1.90).

This will publish all sub-crates linked in the top level workspace
configuration in `Cargo.toml` unless they have `publish = false` defined
in the sub-package.

This also removes the `cargo_publish` output and defers validation to
cargo.

Change-type: minor


pipex:
Testing this here https://github.com/balena-io-modules/mahler-rs/actions/runs/18008752693

pipex:
I'm testing some stuff, I want to see if I can get the binary name with cargo somehow

dfunckt:
> Another ugly workaround would be to assume the github repository name is the package name in the case of rust projects

Yeah I thought about it. It isn't as bad as the other workarounds but I think it should be a last resort.

pipex:
> Try cargo metadata --no-deps --format-version 1 | jq

This contains, which could be an option, but I'm hoping there is an easier way

```
"targets": [
        {
          "kind": [
            "bin"
          ],
          "crate_types": [
            "bin"
          ],
          "name": "helios",
          "src_path": "/Users/felipe/Development/balena/helios/dfunckt/workspace/helios/src/main.rs",
          "edition": "2024",
          "doc": true,
          "doctest": false,
          "test": true
        }
      ],
```

pipex:
Ok, made the change in a new commit. Let's see if the tests pass now

dfunckt (approved):
Nice!

PR #1581, opened 2025-09-29
Always post deploy notes even for unedited release notes

klutchell:
This change is to ensure we don't filter release content based
on whether a human has reviewed the release notes content or not.

If the repo has enabled the release_notes input for an environment,
that is enough to toggle the tag and release notes generation for all
merges as long as there are notable changes detected.

This currently only impacts two internal repositories.

See: https://balena.fibery.io/Work/Improvement/Push-deploy-notes-for-all-production-merges-3267
See: https://balena.fibery.io/Security/Information_Security_and_Reliability_Procedure/Deploy-a-component-to-production-312

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1434,7 +1434,7 @@ jobs:
                 .map(line => `* ${line.replace(/^> -|^-/, '')}`)
                 .join('\n');
 
-              const releaseNotesComment = `#release-notes ${pr.title}\n\nNotable changes\n* [only keep the important and rephrase, leaving this in place will avoid posting release notes]\n${notableChangesFormatted}\n\n${releaseNotesBody}`;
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -1407,7 +1407,7 @@ jobs:
             }
 
             // Check if PR is from Renovate
-            const isRenovate = pr.user.type === 'Bot' && pr.user.login.includes('renovate');
+            const isRenovate = (pr.labels || []).some(label => ((typeof label === 'string' ? label : label.name) || '').toLowerCase() === 'renovate');
```
Page-, line 1410: This can be somewhat simplified by using optional chaining:
```suggestion
            const isRenovate = pr.labels?.some(label => (typeof label === 'string' ? label : label.name)?.toLowerCase() === 'renovate') === true;
```

PR #1763, opened 2026-05-11
Replace check-engine with inline semver for Node 26 check

### LINKED ISSUE #1762: `check-engine` fails on Node v26
The `npx check-engine@1` step in the Node.js workflow fails under Node 26 because the version of yargs bundled by check-engine uses CommonJS require() from an ESM scope, which Node 26 no longer permits:

```
++ node --version
+ node_version=v26.1.0
+ npx -q -y -- check-engine@1
file:///home/runner/.npm/_npx/2f24c27b6b501c67/node_modules/yargs/yargs:3
const {applyExtends, cjsPlatformShim, Parser, Yargs, processArgv} = require('./build/index.cjs')
                                                                    ^

ReferenceError: require is not defined in ES module scope, you can use import instead
    at file:///home/runner/.npm/_npx/2f24c27b6b501c67/node_modules/yargs/yargs:3:69
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:543:37)
    at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:353:47)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1675:24)
    at Module._compile (node:internal/modules/cjs/loader:1838:5)
    at Object..js (node:internal/modules/cjs/loader:1996:10)
    at Module.load (node:internal/modules/cjs/loader:1579:32)
    at Module._load (node:internal/modules/cjs/loader:1381:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1602:12)

Node.js v26.1.0
```

Ref: https://github.com/balena-io/balena-cli/actions/runs/25692837300/job/75433597072?pr=3096

It looks like `check-engine`'s latest release was in Sept 2024: https://github.com/mohlsen/check-engine

Suggestion: drop `check-engine` and move to an inline node script using semver read package.json. Run `semver.satisfies(process.version, engines.node)` directly.

cywang117:
check-engine@1 bundles an old yargs whose unextensioned entrypoint Node 26 loads as ESM, which then throws on its internal require() call. The package has not seen updates since 2024-09-24. For the Node 26 step only, do the engines.node check inline with `node -p` + the semver CLI.

Closes: https://github.com/product-os/flowzone/issues/1762
Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -2113,8 +2113,16 @@ jobs:
           # renovate: datasource=node-version depName=node-26 packageName=node-26.x
           NODE_VERSION: 26.1.0
 
+      # check-engine@1 does not work with Node 26 due to ESM/CommonJS issues
       - <<: *checkEngine
         id: check_engine_26
+        run: |
+          node_version="$(node --version)"
+          range="$(node -p 'require("./package.json").engines?.node || ""')"
+          if [ -z "$range" ] || npx -q -y -- semver -r "$range" "$node_version" >/dev/null
+          then
+            echo "node_version=${node_version}" >> "${GITHUB_OUTPUT}"
+          fi
 
       # Fails if no versions were matched
       # https://github.com/actions/github-script
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -1592,8 +1592,8 @@ jobs:
         id: check_engine_26
         run: |
           node_version="$(node --version)"
-          # If checking the engine passes, or if it fails outright (ie no engines specified) then we accept this version
-          if npx -q -y -- check-engine@1 || ! npx -q -y -- check-engine@1 --ignore
+          range="$(node -p 'require("./package.json").engines?.node || ""')"
+          if [ -z "$range" ] || npx -q -y -- semver -r "$range" "$node_version" >/dev/null
           then
             echo "node_version=${node_version}" >> "${GITHUB_OUTPUT}"
           fi


flowzone.yml
```diff
@@ -2113,8 +2113,16 @@ jobs:
...
+      # check-engine@1 does not work with Node 26 due to ESM/CommonJS issues
       - <<: *checkEngine
         id: check_engine_26
+        run: |
+          node_version="$(node --version)"
+          range="$(node -p 'require("./package.json").engines?.node || ""')"
+          if [ -z "$range" ] || npx -q -y -- semver -r "$range" "$node_version" >/dev/null
+          then
+            echo "node_version=${node_version}" >> "${GITHUB_OUTPUT}"
+          fi
```
klutchell, line 2125: If we go this route, I would prefer we apply the same logic for all the checks for consistency. Not just node 26 where it's required.

flowzone.yml
```diff
@@ -2113,8 +2113,16 @@ jobs:
           # renovate: datasource=node-version depName=node-26 packageName=node-26.x
           NODE_VERSION: 26.1.0
 
+      # check-engine@1 does not work with Node 26 due to ESM/CommonJS issues
       - <<: *checkEngine
         id: check_engine_26
+        run: |
+          node_version="$(node --version)"
```
klutchell, line 2120: The check-engine package also supports checking npm version among a bunch of others, we would want to support that at a minimum.

klutchell:
Superseded by https://github.com/product-os/flowzone/pull/1764

PR #1764, opened 2026-05-12
Force commonjs for check-engine

### LINKED ISSUE #1762: `check-engine` fails on Node v26

klutchell:
Required for node 26 since check-engine ships with a legacy yargs version.

Fixes: https://github.com/product-os/flowzone/issues/1762

Tested:
- Broken ESM import -> throws error and exits
- No engines object in package.json -> assumes all versions supported
- Engines object in package.json -> correctly maps versions in range

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -2072,16 +2072,29 @@ jobs:
           # renovate: datasource=node-version depName=node packageName=node-20.x
           NODE_VERSION: 20.20.2
 
+      - name: Install check-engine
+        run: |
+          npm install -g check-engine@1.14.0
+          # check-engine bundles yargs with "type": "module"; its extensionless
+          # entrypoint then loads as ESM and breaks on its internal require().
+          yargs_pkg="$(npm root -g)/check-engine/node_modules/yargs/package.json"
+          [ -f "$yargs_pkg" ] || yargs_pkg="$(npm root -g)/yargs/package.json"
+          jq '(.type) = "commonjs"' "$yargs_pkg" > "$yargs_pkg.tmp" && mv "$yargs_pkg.tmp" "$yargs_pkg"
+
       # https://www.npmjs.com/package/check-engine
       - &checkEngine
         name: Check engine
         id: check_engine_20
         run: |
           node_version="$(node --version)"
-          # If checking the engine passes, or if it fails outright (ie no engines specified) then we accept this version
-          if npx -q -y -- check-engine@1 || ! npx -q -y -- check-engine@1 --ignore
+          # If checking the engine passes, or if it fails outright (ie no engines specified) then we accept this version.
+          # Even with the --ignore flag, check-engine will still return non-zero if engines are not specified.
+          if check-engine || (check-engine --ignore || true) | grep -q "No engines found"
           then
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -2072,16 +2072,29 @@ jobs:
...
       - &checkEngine
         name: Check engine
         id: check_engine_20
         run: |
           node_version="$(node --version)"
-          # If checking the engine passes, or if it fails outright (ie no engines specified) then we accept this version
-          if npx -q -y -- check-engine@1 || ! npx -q -y -- check-engine@1 --ignore
+          # If checking the engine passes, or if it fails outright (ie no engines specified) then we accept this version.
+          # Even with the --ignore flag, check-engine will still return non-zero if engines are not specified.
+          if check-engine || check-engine --ignore || (check-engine --ignore || true) | grep -q "No engines found"
```
klutchell:
```suggestion
          if check-engine || (check-engine --ignore || true) | grep -q "No engines found"
```
I think we actually want to drop the middle one? We just need to fix the check for unprovided engines, not actually support checking for mismatched entries.

Page-: This seems valid, although it might be worth coupling with targeting a more specific version of check-engine since strings are more likely to change than anything else (although the package seems pretty stable - I'd expect at most a move to ESM but that's not necessary for it)

Page- (approved):

PR #1814, opened 2026-06-18
feat: pin helm to version 4.2.2

cmfcruz:
Change-type: patch


cmfcruz:
This is a mistake.  Please ignore.

PR #1828, opened 2026-06-24
Support digest pinning for binfmt version

klutchell:
Add an optional currentDigest capture group to the custom regex manager so Renovate can pin and maintain the SHA256 digest alongside the binfmt tag. The tag and digest are captured separately, so the inline versioning regex only ever parses the bare tag. Seed the current index digest.

Change-type: patch

---

Should resolve this error preventing the pinDependencies branch from being created
image


pipex (approved):
LGTM

Issue #1855, opened 2026-07-22
Eliminate the `pull_request_target` trust boundary for external PRs


klutchell:
Flowzone grants secrets to external (fork) PRs using `pull_request_target` with a review-gate. actions/checkout v7 now refuses to check out fork PR code under `pull_request_target` (and `workflow_run`) unless you set `allow-unsafe-pr-checkout: true` ([actions/checkout#2454](https://github.com/actions/checkout/pull/2454)). Having to write "unsafe" into our own config is upstream telling us, in plain language, that the pattern is dangerous.

## Problem

The security boundary is procedural, not structural. Untrusted fork code runs in the base repo's trusted context (its `GITHUB_TOKEN`, secrets, default-branch cache scope, and runner access), and the only thing between a fork PR and our registry/cloud/deploy credentials is a reviewer noticing a bad ref, a misplaced secret reference, or a typo before clicking approve. GitHub's own security team documents this as the "pwn request" anti-pattern (see <https://gh.io/securely-using-pull_request_target>). It's secure by policy, and it relies on every Flowzone reviewer being a security expert on every PR.

Full context, appetite, and reach are in the Fibery project: <https://balena.fibery.io/Work/Project/1398> (Shaping, currently paused).

## Direction

Make the boundary structural so untrusted code physically can't reach secrets. Rough shape, still needs shaping:

- Fork PRs run build + test with **no secrets** via the `pull_request` trigger. GitHub's "Approve and run" UI becomes the only gate, which is its intended model.
- Split the steps that need secrets (publish/finalize) out of the untrusted path.
- Surface publish-path failures during PR review (e.g. draft publish via `workflow_run`) so a broken finalize is caught while someone is still watching, instead of weeks later post-merge.

## Related

- Fibery: [Flowzone Security Rearchitecture](https://balena.fibery.io/Work/Project/Flowzone-Security-Rearchitecture-1398)
- [actions/checkout#2454](https://github.com/actions/checkout/pull/2454) (the v7 change) and our bump in #1824
- #515 add external contribution e2e test
- #332 avoid leaking secrets in arbitrary code
- #333 allow external contributions
- Zulip: <https://balena.zulipchat.com/#narrow/channel/348930-balena-io.2Fflowzone/topic/Truffelhog.20sniffs.20through.20flowzone/near/555597293>
[...]

PR #1856, opened 2026-07-22
Flag unsafe fork PR checkout required by checkout v7

klutchell:
checkout v7 refuses to check out fork PR code under pull_request_target unless allow-unsafe-pr-checkout is set. Set it on the three fork-code checkouts to stay on v7, and mark each with a FIXME pointing at the rearchitecture tracking issue so the trust boundary gets fixed properly.

Change-type: patch
See: https://github.com/product-os/flowzone/issues/1855

> **Note**
> There is a real risk that enabling this flag will expose us to more AI and bot pwn targeted attacks as scanners will see it and target repos using it. Being a shared workflow helps a bit, but any AI will quickly determine that all of our repos use this shared workflow.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -458,6 +458,7 @@ jobs:
           fetch-depth: 0
           submodules: recursive
           ref: ${{ github.event.pull_request.head.sha || '¯\_(ツ)_/¯' }}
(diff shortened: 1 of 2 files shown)


klutchell:
Superseded by https://github.com/product-os/flowzone/pull/1862

PR #1862, opened 2026-07-24
Never run fork PRs with secrets (trust boundary Phase 1)

klutchell:
Reworks how Flowzone handles external contributions so **fork PRs never run with secrets**, replacing the `pull_request_target` + review-gate "pwn request" pattern where untrusted fork code ran in the base repo's trusted context.

Design and rationale: [`docs/trust-boundary.md`](./docs/trust-boundary.md)

See: https://github.com/product-os/flowzone/issues/1855
See: https://balena.fibery.io/Work/Project/1398

## Manual testing

Validated the full fork lifecycle on a throwaway repo pinning this branch, forked to a personal account and driven through a real fork PR → merge:

- [x] **Fork PR (no secrets)**: builds and tests; publish/finalize jobs skip; `All jobs` is green. Forks never receive secrets and can't publish.
- [x] **Fork merge → Docker**: the push lane rebuilds from the merged commit and publishes to ghcr (`build-<sha>` + retag to the branch tag) — same-run rebuild, never the fork's uploaded artifact.
- [x] **Fork merge → versioning/release**: cuts the version tag + GitHub release, and the app token writes the versioned commit to the protected default branch as `vX.Y.Z [skip ci]`. Confirmed no recursive run fires for that commit — the `[skip ci]` guard holds.
- [x] **Fork PR footer validation**: a fork PR missing a `Change-type` footer now fails `versioned_source` at PR time ("No commits were annotated with a change type"), instead of only failing after merge. Writes stay trusted-only, so nothing is pushed from the fork.
- [x] **Direct push (no PR)**: quiet skip — no publish or finalize.
- [x] **Caller permissions**: a caller missing `pull-requests: read` / `packages: write` fails the reusable-workflow call at startup, confirming the usage snippet's permission block is required.

### Model

- **Fork PRs** run on `pull_request` with no secrets — build and test only. No approval gate, no `pull_request_target`.
- **Fork publishing** moves to the `push` on the default branch after merge, rebuilt from the merged (trusted) commit — never the fork's uploaded bytes.
- **Internal PRs** are unchanged (secrets on `pull_request`, draft on open, finalize on merge).

The lane is decided in `event_types` and gated per job via `trusted` / `fork_merge` / `push_finalize` outputs. `pull_request_target` is rejected at the gate; its `all_jobs` uses a per-event, non-required check name so a misconfigured caller can't game the required check.

### Breaking changes

- **Fork testing** requires a caller change. The previous caller snippet routed fork PRs to
  `pull_request_target` with an `if:` condition, so a fork's `pull_request` invocation was
  filtered out. Flowzone no longer runs on `pull_request_target`, so callers must adopt the new
  [usage](../README.md#usage) snippet — removing the `pull_request_target` trigger and its
  routing `if:` — to send fork PRs to the no-secrets `pull_request` lane.
- **Fork publishing** additionally requires the `push` trigger on the default branch.
- **Internal PRs** are unchanged. A caller that keeps the old `pull_request` +
  `pull_request_target` wiring keeps working for internal PRs: the `pull_request_target`
  invocation is filtered out by the caller `if:` (and if it does run, Flowzone rejects it and
  its `all_jobs` is the non-required per-event context, so it never gates). Removing the dead
  trigger is otherwise a mechanical follow-up.
- **Caller permissions.** Most callers need no permission change. A caller that explicitly pins
  a restrictive `permissions:` block must include `pull-requests: read` — `event_types`
  declares it for the push-merge PR lookup, and a reusable-workflow call fails at startup if the
  caller granted less. Callers that do not pin the permission inherit enough by default.

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1057,9 +1057,7 @@ jobs:
   event_types:
     name: Event Types
     runs-on: ubuntu-24.04
-    # Wait up to 60 min for workflow approvals on forks.
-    # App Installation tokens expire in 1h either way.
-    timeout-minutes: 60
+    timeout-minutes: ${{ fromJSON(inputs.jobs_timeout_minutes) }}
     if: |
       (
         (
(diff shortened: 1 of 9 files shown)


flowzone.yml
```diff
@@ -709,11 +656,6 @@
     continue-on-error: true
     run: aws sts get-caller-identity
 
-  - &updateKubeconfig
```
ab77, line 712: note: not used

ab77 (approved):
@klutchell my only suggestion would be to run `actionlint .github/workflows/flowzone.yml` and address mostly cosmetic `shellcheck` suggestions (which we never enforce).

PR #1868, opened 2026-07-29
Do not require pull-requests: read from callers

klutchell:
`event_types` declared a job-level permissions block with `pull-requests: read`, which made the reusable workflow require that scope of every caller. Callers whose org default GITHUB_TOKEN omits pull-requests fail the permissions subset check at startup, breaking all of their Flowzone runs.

Drop the permissions block so event_types inherits the caller's token. The push-lane fork-merge PR lookup still needs pull-requests: read, so mint an ephemeral Flowzone app token scoped to it.
A push is trusted, so `app_id` and the app key are available.

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1035,10 +1035,10 @@ jobs:
           - event_name: ${{ github.event_name }}
             event_action: ${{ github.event.action }}
 
-    permissions:
-      # Read-only PR lookup for `Classify push merge` below (uses github.token on the push lane).
-      pull-requests: read
-
+    # No job-level permissions block: declaring pull-requests: read here made the reusable
+    # workflow require it of every caller, and callers whose org default omits it fail the
+    # permissions subset check at startup. event_types instead inherits the caller's token; the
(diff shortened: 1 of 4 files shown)


ab77 (approved):

flowzone.yml
```diff
@@ -1056,11 +1069,21 @@ jobs:
...
-              commit_sha: context.sha,
-            });
+            let prs;
+            try {
+              ({ data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+                ...context.repo,
+                commit_sha: context.sha,
+              }));
+            } catch (e) {
+              core.warning(`Could not look up the PR for this push (needs 'pull-requests: read'): ${e.message}. Fork-merge publishing will be skipped.`);
```
pipex, line 1084: What does it mean that `fork-merge` publishing is skipped? PRs from contributions won't be able to merge to main?

klutchell: They'll merge, they just won't publish anything.

PR #1871, opened 2026-07-29
Stop describing the push trigger as optional

klutchell:
Replaced with a description of what the trigger is for, and nothing about whether it's needed:

```yaml
  # Fork contributions are rebuilt and published from the push to the default
  # branch after merge.
  push:
    branches: [main, master]
```

The matching "Accepted limitations" bullet in `docs/trust-boundary.md` now notes the trigger will be needed, rather than framing its absence as a settled choice.

---

Also keep 'skip ci' out of the version commit subject


kb2ma:
Does this change mean that in an example like below the first commit now will show an ellipsis like the second commit?


image

pipex (approved):

PR #1878, opened 2026-08-10
Stop suppressing every workflow on the version commit

klutchell:
The version commit that versioned_source pushes carried [skip ci]. GitHub applies skip directives repository-wide rather than per-workflow, so it also silenced downstream pipelines that key deployments off pushes to the default branch (balena-os).

Replace the directive with a Flowzone-version-commit trailer and veto the re-entrant run inside Flowzone instead, so only Flowzone skips. Every job lists event_types in its needs, directly or transitively, which makes that job's if: the single point able to veto a whole run. Two independent signals, because either can be absent: the app's sender login covers any push the app makes, and the trailer covers the legacy FLOWZONE_TOKEN path where the pusher is a human account.

The veto is required rather than a cost saving. balena_publish, website_publish and custom_always gate only on `trusted`, which is true for any push, so a self-triggered run would push a second balena release and redeploy the Cloudflare Pages site.

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -270,7 +270,9 @@ jobs:
         github.event_name == 'push' && (
           startsWith(github.ref, 'refs/tags/') ||
           github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
-        )
+        ) &&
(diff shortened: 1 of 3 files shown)


shaunmulligan (approved):
LGTM

PR #1890, opened 2026-08-19
Add a job that warns when the caller workflow needs migrating

klutchell:
Flowzone's caller contract changes over time, and a drifted caller gets no
signal today: nothing fails, features quietly stop applying. check_config runs
flowzonify against the caller's own flowzone.yml and warns per outstanding
migration. It holds no token, never gates a merge, and fixing it stays the
maintainer's move.

See: https://balena.fibery.io/Work/Improvement/Migration-PRs-open-themselves-4663

e.g.
image

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1929,6 +1929,178 @@ jobs:
               sarif: zippedSarif,
             });
 
+  # Report, but never fix, outstanding migrations to this repository's own Flowzone caller
+  # workflow. The caller contract changes over time, and a caller that has drifted gets no
+  # other signal: nothing fails, features quietly stop applying.
+  #
+  # Deliberately absent from `all_jobs` needs: a drifted caller workflow is not a reason to
+  # block the change under review.
+  check_config:
+    name: Check config
+    runs-on: ${{ fromJSON(inputs.runs_on) }}
+    timeout-minutes: 5
+    needs:
+      - event_types
+
+    # Trusted lanes only. Not for secrecy -- there is no secret here -- but because the step
+    # below resolves an npm dependency tree inside the checked-out working copy, which is
+    # attacker-controlled on a fork pull request. Nothing is lost: a fork contributor does not
+    # own the caller workflow and cannot act on the warning.
+    #
+    # Closed pull requests are skipped because nobody acts on that report, and tag pushes
+    # because they duplicate the default-branch push that produced the tag.
+    if: |
+      needs.event_types.outputs.trusted == 'true' &&
+      github.event.action != 'closed' &&
+      !startsWith(github.ref, 'refs/tags/')
+
+    # Only what actions/checkout needs to clone a private caller. Nothing here writes.
+    permissions:
+      contents: read
+
+    # Named here so the step that writes them and the step that reads them share one owner.
+    # `runner.temp` is not available to a job-level `env`, so the directory is joined at use.
+    env:
+      LISTING_NAME: flowzonify-list.txt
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -1929,6 +1929,204 @@ jobs:
...
+
+            // `--list` prints each migration as an unindented id followed by its description
+            // indented on the next line. Parsed leniently and never fatally: if that format
+            // changes, the warnings fall back to bare ids rather than breaking.
+            const described = new Map();
+            try {
+              const lines = fs.readFileSync(process.env.LISTING, "utf8").split("\n");
+              lines.forEach((line, index) => {
+                const next = lines[index + 1] || "";
+                if (line && !line.startsWith(" ") && next.startsWith("  ")) {
```
thgreasi: Would it work if we checked that it starts w/ a number as in the comment?
```suggestion
                if (line && line.match(/^\d/) && next.startsWith("  ")) {
```

flowzone.yml
```diff
@@ -1929,6 +1929,204 @@ jobs:
...
+              core.warning(
+                "Could not check this repository's Flowzone caller workflow: " + error.message,
+              );
+              return;
+            }
+
+            // Every warning below stands on its own -- what is outstanding and how to fix
+            // it -- because GitHub renders annotations as a flat list interleaved with other
+            // jobs', where a separate summary reads as unrelated rather than as a total.
+            const FIX = "Run `npx product-os/flowzonify` to apply it.";
```
thgreasi: Should it be v ?
```suggestion
            const FIX = "Run `npx flowzonify` to apply it.";
```

flowzone.yml
```diff
@@ -1929,6 +1929,204 @@ jobs:
...
+              // `changed` is false here, so it is reported but never counted as migratable.
+              if (file.status === "blocked") {
+                for (const unit of file.report ?? []) {
+                  if (unit.status === "blocked") {
+                    core.warning(
+                      `${unit.id} needs a manual migration: ${unit.note || "no detail was reported"}`,
+                      anchor(file.path),
+                    );
+                  }
+                }
```
thgreasi: NIT since it seems that we expect `file.report` to alway be populated, so if it isn't, we better raise it.
```suggestion
                if (file.report == null) {
                  core.warning(
                      `${file.path} needs a manual migration: no detail was reported`,
                      anchor(file.path),
                    );
                    continue;
                }
                for (const unit of file.report) {
                  if (unit.status === "blocked") {
                    core.warning(
                      `${unit.id} needs a manual migration: ${unit.note || "no detail was reported"}`,
                      anchor(file.path),
                    );
                  }
                }
```

flowzone.yml
```diff
@@ -1929,6 +1929,204 @@ jobs:
...
+                if (unit.status === "skip") {
+                  continue;
+                }
+                // `note` is only set for an advisory, and is more specific than the
+                // migration's general description when it is there.
+                const why = unit.note || described.get(unit.id);
+                // Punctuation normalised so the fix reads as its own sentence whatever the
+                // prose ends on, and the id reads as a subject when there is no prose at all.
+                const problem = why
+                  ? `${unit.id}: ${why.endsWith(".") ? why : `${why}.`}`
```
thgreasi: NIT: I would probably move the `.` suffix in a `if(why && !why.endsWith(".")) { why += '.' }` right above, and avoid the nested ternaries, but that's not a blocker :)

thgreasi (approved):

PR #1896, opened 2026-08-24
Skip publish jobs on internal merge pushes

cywang117:
balena_publish, docker_publish and website_publish had no push-lane gate, so a merge to the default branch published a duplicate release alongside the pull_request lane. Gate them on fork_merge, matching custom_publish.

Change-type: patch

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -1014,9 +1014,9 @@ jobs:
     #
     # The push lane excludes Flowzone's own pushes. The versioned commit is pushed with the app
     # token, which (unlike github.token) does re-trigger workflows, so it re-enters the pipeline.
-    # That run cannot loop (no merged PR -> fork_merge and push_finalize are false), but the jobs
-    # with no push-lane gate beyond `trusted` (balena_publish, website_publish, custom_always)
-    # would publish a second time. Two independent signals, because either can be absent:
+    # That run cannot loop (no merged PR -> fork_merge and push_finalize are false), but any job
+    # with no push-lane gate beyond `trusted` (custom_always) would publish a second time. Two
+    # independent signals, because either can be absent:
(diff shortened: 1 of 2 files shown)


flowzone.yml
```diff
@@ -3823,7 +3823,8 @@ jobs:
       needs.event_types.outputs.trusted == 'true' && (
       !failure() && !cancelled() &&
       needs.docker_test.result == 'success' &&
-      needs.is_docker.outputs.docker_publish_matrix != ''
+      needs.is_docker.outputs.docker_publish_matrix != '' &&
+      (github.event_name != 'push' || needs.event_types.outputs.fork_merge == 'true')
```
klutchell, line 3827: This one is no-op as `docker_test` only runs on an open PR or a fork-merge push. It doesn't hurt, but it was never running on internal PR pushes.

flowzone.yml
```diff
@@ -4044,7 +4045,8 @@ jobs:
       needs.event_types.outputs.trusted == 'true' && (
       !failure() && !cancelled() &&
       needs.is_balena.result == 'success' &&
-      (github.event.action != 'closed' || github.event.pull_request.merged == true)
+      (github.event.action != 'closed' || github.event.pull_request.merged == true) &&
+      (github.event_name != 'push' || needs.event_types.outputs.fork_merge == 'true')
```
klutchell, line 4049: Unfortunately this also skips push events where versioning is disabled and Flowzone is intended to run on tag events. We only want to block the internal-merge push, so something like this would work instead

```yaml
(github.event_name != 'push'
     || needs.event_types.outputs.fork_merge == 'true'
     || needs.event_types.outputs.merged_pr == '')
```

klutchell (changes requested):

klutchell:
Superseded by https://github.com/product-os/flowzone/pull/1899

PR #1905, opened 2026-08-27
Verify submodule commits are on their declared .gitmodules branch

guille-vega:
Some device repositories don't declare a branch in .gitmodules. When it's missing, Renovate doesn't know which branch we actually meant to track, and ends up following the submodule's default branch instead rather than the branch it's actually pinned to. So Renovate opens PR updates to latest default instead of the intended branch.

Add an opt-in check_gitmodules_branches input that fails if a submodule has no declared branch, or its pinned commit isn't on it. Lives inside versioned_source, right after the submodule checkout it already does.

Change-type: minor

### DIFF
--- a/flowzone.yml
+++ b/flowzone.yml
@@ -959,6 +959,11 @@ on:
         type: boolean
         required: false
         default: true
+      check_gitmodules_branches:
+        description: "Require every .gitmodules submodule to declare a branch, and its pinned commit to be on it."
+        type: boolean
+        required: false
+        default: "${{ github.repository_owner == 'balena-os' }}"
     outputs:
       cloudflare_deployment_url:
         description: "Cloudflare Deployment URL"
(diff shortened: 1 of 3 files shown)


flowzone.yml
```diff
@@ -959,6 +959,11 @@ on:
         type: boolean
         required: false
         default: true
+      check_gitmodules_branches:
+        description: "Require every .gitmodules submodule to declare a branch, and its pinned commit to be on it."
+        type: boolean
+        required: false
+        default: false
```
klutchell:
```suggestion
        default: ${{ github.owner == 'balena-os' }}
```
Or maybe even `${{ github.event.enterprise.slug == 'balena' }}` if we want to auto-enable it internally for all orgs, because we use renovate everywhere.

flowzone.yml
```diff
@@ -1262,6 +1267,45 @@ jobs:
           ref: ${{ github.sha || '¯\_(ツ)_/¯' }}
           <<: *checkoutAuth
 
+      # Require every submodule to declare a branch, and its pinned commit to be on it.
+      # Submodules were already fetched with full history above, so no extra fetch is needed.
+      - name: Check submodule branches
```
klutchell, line 1274:
```suggestion
      - name: Check submodule branches
         id: check_submodules
```
We also need to add this to the `all_jobs` needs array, otherwise a failure could be ignored and PRs merged anyway

flowzone.yml
```diff
@@ -1262,6 +1267,45 @@ jobs:
           ref: ${{ github.sha || '¯\_(ツ)_/¯' }}
           <<: *checkoutAuth
 
+      # Require every submodule to declare a branch, and its pinned commit to be on it.
+      # Submodules were already fetched with full history above, so no extra fetch is needed.
```
klutchell, line 1273: Without getting too specific, we could still comment that as an example tools like Renovate can only manage submodules properly when they know the correct branch. Again, without directly pointing to our use cases, but just as an example.

flowzone.yml
guille-vega: Sounds good to me, but we'll have to fix all repos that don't specify a branch before merging this PR.

flowzone.yml
guille-vega: This is a step inside versioned_source (already a required job for all_jobs), so if it fails, versioned_source fails and it'll block merge.

flowzone.yml
klutchell: Or use this change as a forcing function

flowzone.yml
klutchell: Right, my mistake.

klutchell:
@guille-vega you'll need to rebase locally and `npm run build` to resolve the conflict as the actual workflow file is generated as a post-commit hook.

klutchell (approved):

PR #1912, opened 2026-09-06
Make .gitmodules check opt-in for balena-os

klutchell:
There are still too many submodules tracking default branches as intended that are not declared and have no requirement for this declaration to start enforcing it org-wide.

In the meantime, we can start opening PRs to declare all submodule branches without otherwise further blocking the pipelines.

Related-to: https://github.com/product-os/flowzone/pull/1905
Change-type: minor


guille-vega (approved):
LGTM

